PR #2, opened 2021-05-17
Featurereleng 4473 main logic

erwan-b:


runners-manager/vm_creation/openstack.py
```diff
@@ -0,0 +1,80 @@
...
+        auth_url=keystone_endpoint,
+        token=token,
+        project_id=tenant_id)
+)
+nova_client = client.Client(version=2, session=session, region_name=region)
+neutron = neutronclient.v2_0.client.Client(session=session, region_name=region)
+
+
+def script_init_runner(name, token, labels, group):
+    installer = link_download_runner('scalanga-devl')
```
tcarmet, line 30: something to keep in mind, but we need a variable here before I can actually set something up deployed

runners-manager/vm_creation/openstack.py
```diff
@@ -0,0 +1,80 @@
...
+    :return:
+    """
+    sec_group_id = neutron.list_security_groups()['security_groups'][0]['id']
+    nic = {'net-id': neutron.list_networks(name='tenantnetwork1')['networks'][0]['id']}
+    print("creating virtual machine")
+    print("----------------------------")
+    # print(script_init_runner(name, runner_token, ['centos'], 'default'))
+    # print("----------------------------")
+    instance = nova_client.servers.create(
+        name=name, image='dcfbc223-d658-4077-a517-5f29984258a6',
```
tcarmet, line 62: can we use image name instead of IDs? Might be more readable.

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,11 @@
...
+
+on: push
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - run: exit 0
```
tcarmet, line 11: Could you just add a `flake8` command? So that at least we lint the code from day 1

tcarmet (approved):
LGTM, left some comments that are up to you to either adress now or later. (but keep them tracked)

PR #3, opened 2021-05-19
[releng-4489] Create docker and helm

erwan-b:


charts/openstack-actions-runner/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
...
+    {{- include "openstack-actions-runner.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": test
+spec:
+  containers:
+    - name: wget
+      image: busybox
+      command: ['wget']
+      args: ['{{ include "openstack-actions-runner.fullname" . }}:{{ .Values.service.port }}']
+  restartPolicy: Never
```
tcarmet, line 15: We can remove this file

runners-manager/vm_creation/github_actions_api.py
```diff
@@ -6,54 +6,49 @@
 
 HEADERS = {
     'Accept': 'application/vnd.github.v3+json',
-    'Authorization': f'token {os.getenv("GITHUB_TOKEN")}'
+    'Authorization': f'token {os.getenv("githubToken")}'
```
tcarmet, line 9: I made a mistake expressing myself about the camelcase. I meant that it should be only on the helm chart, no need of doing it in the actual code.

Because we usually see upper case when using env variable, and that is probably the direction we are going to move forward with.

charts/openstack-actions-runner/templates/_helpers.tpl
```diff
@@ -0,0 +1,75 @@
...
+*/}}
+{{- define "helpers.list-env-variables"}}
+{{- range $key, $val := .Values.env.secret }}
+- name: {{ $key }}
+  valueFrom:
+    secretKeyRef:
+      name: app-env-secret
+      key: {{ $key }}
+{{- end }}
+{{- end }}
```
tcarmet, line 75: missing \n

tcarmet (approved):
LGTM

PR #4, opened 2021-06-01
[releng-4537] Manage multi vm flavor images

erwan-b:


runners-manager/main.py
```diff
@@ -1,114 +1,47 @@
...
-            'parent_name': None
-        }
-
-        runner_counter += 1
-
-    return runner_counter, runner_infos
-
-
 def main():
     org = 'scalanga-devl'
```
tcarmet, line 27: TODO: create a variable for that

runners-manager/runner/RunnerManager.py
```diff
@@ -0,0 +1,76 @@
...
+    runners: dict[str, Runner]
+    runner_management: list[VmType]
+
+    def __init__(self, org: str, config: list):
+        self.runner_counter = 0
+        self.github_organization = org
+        self.runner_management = [VmType(elem) for elem in config]
+        self.runners = {}
+
+        for t in self.runner_management:
```
tcarmet, line 20: Some thoughts to be noted. We shoudn't delete the runners when we reboot the service. And we don't necessarily need to create them when we boot runner manager.

We should:
* Do nothing when we reboot runner manager
* Find created VMs when we boot the runner
* Create new one if needed

tcarmet: @erwan-b your task is to create a ticket for this comment :up:

runners-manager/vm_creation/openstack.py
```diff
@@ -25,12 +26,36 @@
...
-def script_init_runner(name, token, labels, group):
-    installer = link_download_runner('scalanga-devl')
-    return f"""#!/bin/bash
+def install_docker(image: str):
+    if 'centos' in image.lower():
+        return """
 sudo yum install -y bind-utils yum-utils
 sudo yum-config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
 sudo yum install -y epel-release docker-ce docker-ce-cli containerd.io
+""" # noqa
```
tcarmet, line 35: Coment for all the scripts:
* Create separate file for all of them that you open and use
* If needed you can add jinja template

tcarmet (approved):

PR #5, opened 2021-06-04
[RELENG-4538] add scalability on waiting vm

erwan-b:


runners-manager/main.py
```diff
@@ -17,34 +18,32 @@ def maintain_number_of_runner(runner_m: RunnerManager):
             f"offline: {len([e for e in runners['runners'] if e['status'] == 'offline'])}"
         )
 
-        for elem in runners['runners']:
-            runner_m.update_runner(elem)
+        print(runners)
```
tcarmet, line 21: We need to start to think about adding a real logger instead of using print statements.

runners-manager/main.py
```diff
@@ -17,34 +18,32 @@ def maintain_number_of_runner(runner_m: RunnerManager):
...
-        'quantity': 2,
-    }, {
-        'tags': ['ubuntu', '16.04', 'small'],
-        'flavor': 'm1.small',
-        'image': 'Ubuntu 16.04 LTS (Xenial Xerus) (PVHVM)',
-        'quantity': 2,
+        'quantity': {
+            'min': 2,
+            'max': 4,
+        },
```
tcarmet, line 35: I like the concept of having min max, good idea.

tcarmet (approved):
Few lines of code for such a feature, I like, well done

PR #6, opened 2021-06-07
Add password connection

erwan-b:


tcarmet (approved):

charts/openstack-actions-runner/Chart.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: v2
+apiVersion: v1
```
tcarmet, line 1: I would recommend just removing this field. (if that works)

That's our bad but currently we use helm2 in releng-it and you are using helm3 (and you should). But this chart needs to be compatible with both for now.

erwan-b: It don't work without the `apiVersion`

PR #12, opened 2021-06-15
[RELENG-4601] add catch error for error from API

erwan-b:
Remove docker installation to speed up virtual machine setup. And add security around github API call

Tests with docker installation and simple command: https://github.com/scalanga-devl/k8s-intro/runs/2829634223?check_suite_focus=true
Actions with matrix: https://github.com/scalanga-devl/k8s-intro/actions/runs/938584973
thing we must avoid: https://github.com/scalanga-devl/k8s-intro/actions/runs/939461884

### DIFF
--- a/runners_manager/main.py
+++ b/runners_manager/main.py
@@ -1,7 +1,7 @@
 import time
 import logging
 
-from vm_creation.github_actions_api import get_runners, force_delete_runner
+from vm_creation.github_actions_api import get_runners
 from runner.RunnerManager import RunnerManager
 
 logger = logging.getLogger("runner_manager")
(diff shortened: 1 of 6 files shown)


runners_manager/main.py
```diff
@@ -25,8 +25,11 @@ def main(org):
     runners = get_runners(org)
     logger.debug("Delete all runners")
     logger.debug(runners)
-    for elem in runners['runners']:
-        force_delete_runner(org, elem['id'])
+    # for elem in runners['runners']:
+    #     try:
+    #         force_delete_runner(org, elem['id'])
+    #     except APIException:
+    #         print('API exception')
```
tcarmet, line 32: Either remove the code or put it back, but let's not leave commented code

tcarmet (approved):
LGTM
Just please address my one comment in the way you want.

PR #14, opened 2021-06-17
Use the right script in the dockerfile

erwan-b:


Dockerfile
```diff
@@ -33,4 +33,4 @@ RUN pip3 install -r ./requirements.txt
 
 COPY runners_manager /app
 
-CMD python3 ./main.py
+CMD python3 ./start.py
```
tcarmet, line 36: We're gonna have to consider installing the python package inside the docker image to then use it in as the "binary" instead of calling the python file

tcarmet (approved):

Dockerfile
erwan-b: it is T_T

PR #17, opened 2021-06-18
[RELENG-4593] Add config file

erwan-b:


runners_manager/main.py
```diff
@@ -21,15 +24,16 @@ def maintain_number_of_runner(runner_m: RunnerManager):
...
-        'flavor': 'm1.small',
-        'image': 'CentOS 7 (PVHVM)',
-        'quantity': {
-            'min': 2,
-            'max': 4,
-        },
-    }]
-    runner_m = RunnerManager(org, config)
+def main(settings: dict, args: argparse.Namespace):
+    print(args)
```
tcarmet, line 28: use logger

runners_manager/main.py
```diff
@@ -21,15 +24,16 @@ def maintain_number_of_runner(runner_m: RunnerManager):
...
+    importlib.import_module(settings['python_config'])
```
tcarmet, line 29: What is this? :sweat_smile:

runners_manager/start.py
```diff
@@ -1,14 +1,43 @@
-import os
+import argparse
 
-import runners_manager.settings.settings_local # noqa
+from runners_manager.settings.yaml_config import setup_settings
+# import runners_manager.settings.settings_local # noqa
```
tcarmet, line 4: remove comment

runners_manager/start.py
```diff
@@ -1,14 +1,43 @@
...
+    parser.add_argument(
+        '--cloud_nine_user',
+        help="User for cloud nine connection  (Goes with password)")
+    parser.add_argument(
+        '--cloud_nine_password',
+        help="Password for cloud nine connection (Goes with user)")
+    parser.add_argument(
+        '--cloud_nine_token',
+        help="Token for cloud nine connection")
+    return parser
```
tcarmet, line 29: I would rather see all settings use `-` rather than `_`

runners_manager/start.py
```diff
@@ -1,14 +1,43 @@
...
+                                     description='Start and manage github self-host runners')
+
+    parser.add_argument(
+        'settings_file',
+        default='setting.yaml',
+        help="Path to project settings file")
+    parser.add_argument(
+        'github_token',
+        help="Token for github API calls",
+    )
```
tcarmet, line 19: are we expecting to now set those args through command line args? I would rather continue to see "secret configs" as environment variable.

We can still keep the option of having it in the args but I would suggest something like this:
```suggestion
    parser.add_argument(
        'github_token',
        default=os.getenv('GITHUB_TOKEN'),
        help="Token for github API calls",
    )
```
This argument is for every setting that is or works with a secret.

test.yml
```diff
@@ -0,0 +1,13 @@
...
+python_config: 'runners_manager.settings.settings_local'
+runner_pool:
+  - flavor: 'm1.small'
+    image: 'CentOS 7 (PVHVM)'
+    quantity:
+      min: 2
+      max: 4
+    tags:
+      - centos7
+      - small
```
tcarmet, line 13: You can keep a sample.yml file but please remove any value that represent you, or our field as this repository is public

tcarmet (commented):
First batch of review, I'm still reading this PR though, more might be coming

runners_manager/main.py
erwan-b: the function `import_module` allow the user to put a path to a module in a string and import it.
With this way we can put a path to a config file with the python module syntax and import it easily.

erwan-b: And so avoid empty import for a python setup config file

config_example.yml
```diff
@@ -0,0 +1,28 @@
...
+#  runner_pool:
+#    - flavor: 'm1.small'
+#      image: 'CentOS 7 (PVHVM)'
+#      quantity:
+#        min: 2
+#        max: 4
+#      tags:
+#        - centos7
+#        - small
+runner_pool: []
```
tcarmet, line 28:
```suggestion
runner_pool: []

```
adding a `\n`

tcarmet (commented):
Overall LGTM. Here's the next todo list:
* I don't see any unit test that are specific to the config file, could be nice to add some just to ensure we can load the file, get some values, and ensure we handle errors in a fashioned way. I believe this can be fully tested without mocks.
* The chart was not updated to now contain a configmap that you will attach in the runner environment to configure it. The requirements are that you will create a new template configmap file, and that from the `values.yaml` file the user that will deploy the chart can set all the values that goes inside this file to configure the service.

charts/openstack-actions-runner/templates/configmap.yaml
```diff
@@ -6,3 +6,9 @@ data:
   CLOUD_NINE_REGION:  {{ .Values.cloudNineRegion }}
   GITHUB_ORGANIZATION:   {{ .Values.githubOrganization }}
   CLOUD_NINE_TENANT: {{ .Values.cloudNineTenant }}
+  settings.yml: |-
+    github_organization: {{ .Values.githubOrganization }}
+    cloud_nine_region: {{ .Values.cloudNineRegion }}
+    cloud_nine_tenant: {{ .Values.cloudNineTenant }}
+    python_config: {{ .Values.pythonConfigModule }}
+    runner_pool: {{ .Values.runnerPool }}
```
tcarmet, line 14: You have to set this in a different file as this current configmap is being used to attach environment variables. and `settings.yml` is not an environment variable.

charts/openstack-actions-runner/values.yaml
```diff
@@ -59,10 +59,41 @@ tolerations: []
 affinity: {}
 
 githubToken: ""
-githubOrganization: ""
+githubOrganization: "scalanga-devl"
```
tcarmet, line 62:
```suggestion
githubOrganization: ""
```

charts/openstack-actions-runner/values.yaml
```diff
@@ -59,10 +59,41 @@ tolerations: []
...
 
+# Cloud nine connection infos
 cloudNineTenant: ""
 cloudNineToken: ""
 cloudNineUsername: ""
 cloudNinePassword: ""
-cloudNineRegion: ""
+cloudNineRegion: "Europe"
```
tcarmet, line 69:
```suggestion
cloudNineRegion: ""
```

tcarmet (commented):
Here's the main thing that is missing on the helm chart.

You need to attach the `settings.yml` file in a specific path in the runtime of the container. Therefore you need to update your deployment file to say, hey look for this confimap, and the key `settings.yml` and attach this file into the unix path. So that when your runner will start, it can retrieve the file on the filesystem. Today it's not

tcarmet (commented):
So now we have a new configmap file for the runner, which is great.

But I must insist the issue I raised in my previous comment:
> Therefore you need to update your deployment file to say, hey look for this confimap, and the key settings.yml and attach this file into the unix path. So that when your runner will start, it can retrieve the file on the filesystem. Today it's not

Have a look at this piece of documentation to understand a bit more what I'm trying to explain: https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#populate-a-volume-with-data-stored-in-a-configmap

Or here's an example we have in our codebase https://github.com/scality/pipeline-deploy/blob/development/2.0/salt/states/eve-kube/files/clouddb.yml.jinja#L39 Here you can think about how come we can retrieve the file `creds.json`

charts/openstack-actions-runner/templates/deployment.yaml
```diff
@@ -33,10 +33,14 @@ spec:
...
-              name: {{ .Release.Name }}-config
-          - secretRef:
-              name: {{ .Release.Name }}-auth
+            - configMapRef:
+                name: {{ .Release.Name }}-config
+            - secretRef:
+                name: {{ .Release.Name }}-auth
+          volumeMounts:
+            - name: {{ .Release.Name }}-settings
+              mountPath: /app
```
tcarmet, line 42: is the default path to the settings app on the file `/app/settings.yaml` ? if so then LGTM

tcarmet (approved):
LGTM, now the config is binded onto the deployment, I hope that my comment on the deployment file is correct, otherwise we might need to adjust some last settings to tell the app where is the settings.yaml file

charts/openstack-actions-runner/templates/settings.yml
```diff
@@ -0,0 +1,11 @@
...
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}
+data:
+  settings.yml: |-
+    github_organization: {{ .Values.githubOrganization }}-settings
+    cloud_nine_region: {{ .Values.cloudNineRegion }}
+    cloud_nine_tenant: {{ .Values.cloudNineTenant }}
+    python_config: {{ .Values.pythonConfigModule }}
+    runner_pool: {{ .Values.runnerPool }}
```
tcarmet, line 11: I'm trusting you on the output of this file. I hope that you don't need a setting like `runner_pool: {{ .Values.runnerPool | toYaml }}`

PR #24, opened 2021-06-22
[RELENG-4621] infos about the error with self-hosted --once

erwan-b:


tcarmet (commented):
It's looking good, I added some correction to the english writing mostly, the content underneath LGTM. 

can you please move all the documents into a `docs` folder.

Per example: 
`docs/design/github_flag_once_error.md`
`docs/asset/my-asset.png`
etc etc

github_flag_once_error.md
```diff
@@ -0,0 +1,27 @@
...
+
+In a few words: at the end of the self-hosted life, Github gives him a job to execute, but he stops running just after that. 
+Therefore, the job is never start. That block the runner to be deleted, and the job assigned to it, because Github wait for that job to be executed on this offline runner.
+
+## How to avoid this error
+A solution could be to always use the same runners, but on a fresh VM each time. If we save the self-hosted  root directory and put it on a new VM, Github will consider it as the same runner. 
+And it will run the job that Github attached him just before shut down, and avoid any error.
+
+For infos here are file created by the self-hosted scripted, one after the other, we could use this info to save only config file. But ! These files are not meant to by managed by us, so if Github change anything, it could break the all CI.
+![image](./assets/created_file_durring_installation.png)
```
tcarmet, line 24: So I looked at the image, and I feel like the file that we might need are:
* .credentials
* .credentials_rsparam (something like that :sweat_smile: )
* .runner

Should we list them here? Or are we in a state that we're not sure about this piece of information?

erwan-b: Exactly, this 3 files looks like the config file use by the `self-hosted` to connect it's self to github.com. 
One thing bother me, this files are not in any public documentation, so there can be change at any time, new file, name chancing, etc...
The other thing is about the way the `self-hosed` update him self.

tcarmet: For the undocumented files, indeed, it's a gamble we're taking here.

As per the update, as we will always download the runner, and hopefully the latest one, I don't think we should be affected by this issue.

Hopefully by reaching out to github we'll be able to get their thoughts towards our previous solution and see if we really need this one.

tcarmet (approved):

PR #44, opened 2021-10-05
[RELENG-5000] Add web interface

erwan-b:


.github/workflows/release.yml
```diff
@@ -50,3 +50,41 @@ jobs:
...
+          tags: "registry.scality.com/openstack-actions-runner/web-interface:${{ github.event.inputs.tag }}"
+
+      - name: Create Release
+        uses: softprops/action-gh-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.event.inputs.tag }}
+          release_name: Release ${{ github.event.inputs.tag }}
+          prerelease: ${{ github.event.inputs.prerelease }}
```
tcarmet, line 90: What I don't like here (and I don't even know if it will work) is that you're trying to tag the repository twice.

Here what I suggest you to do:
* Split the workflow in three jobs.
* Two jobs that will be executed at the same time and build each docker image and push it to the registry.
* A third job that will just create the GitHub release. and "needs" the other two job to finish before starting.

Or you can also consider adding everything in the same job and just execute a docker build after the other

tcarmet (changes requested):

.github/workflows/release.yml
erwan-b: image
:+1:

PR #47, opened 2021-10-29
support self-hosted runner at the repo level

tcarmet:
In order to help development, adding support of self-hosted runners
at the repository level.

### DIFF
--- a/charts/openstack-actions-runner/values.yaml
+++ b/charts/openstack-actions-runner/values.yaml
@@ -99,6 +99,7 @@ affinity: {}
 
 githubToken: ""
 githubOrganization: ""
+githubRepo: ""
 
 # Cloud nine connection infos
 cloudNineTenant: ""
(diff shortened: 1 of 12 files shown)


charts/openstack-actions-runner/values.yaml
```diff
@@ -99,6 +99,7 @@ affinity: {}
 
 githubToken: ""
 githubOrganization: ""
+githubRepo: ""
```
erwan-b, line 102: With this setting we will be forced to publish one instance for each repository.
Let's find a way were it's not the cases, it's to restrictive in the long terme.

runners_manager/vm_creation/github_actions_api.py
```diff
@@ -7,27 +7,36 @@
...
+    def __init__(self, organization, repo, token):
         self.organization = organization
+        self.repo = repo
+        if self.repo:
+            self.github_api = \
+                f"https://api.github.com/repos/{self.organization}/{self.repo}"
+        elif self.organization:
+            self.github_api = \
+                f"https://api.github.com/orgs/{self.organization}"
+
```
erwan-b, line 23: This part is fin and should be kept.

erwan-b (commented):
Let's discus about it together

PR #49, opened 2021-11-03
First monitoring setup

tcarmet:


runners_manager/runner/Runner.py
```diff
@@ -80,21 +88,29 @@ def toJson(self):
...
-        else:
-            status = elem['status']
-
+    def update_status(self, status: str):
         if self.status == status:
             return
 
-        if self.status == 'offline' and status != 'offline':
+        if self.status not in ['online', 'running'] and status in ['online', 'running']:
             self.started_at = datetime.datetime.now()
```
erwan-b, line 96: This will create problems
Calling `update_status` 2 time with 'oneline'  or 'running' will overwrite the first `datetime.now()`

This problem was already here before your edit :sweat_smile:

erwan-b: Forget it

runners_manager/runner/RunnerManager.py
```diff
@@ -64,23 +65,33 @@ def update(self, github_runners: list[dict]):
         for elem in github_runners:
             if elem['name'] in self.runners:
                 runner = self.runners[elem['name']]
-                runner.update_status(elem)
-
+                if elem['status'] == 'online' and elem['busy'] is True:
+                    status = 'running'
+                else:
+                    status = elem['status']
+                runner.update_status(status)
```
erwan-b, line 72: I don't like the idea  that a `Runner` state is calculated out of it

runners_manager/metrics/Runner.py
```diff
@@ -0,0 +1,11 @@
...
+
+from prometheus_client import Enum
+
+
+PROMETHEUS_RUNNER_STATUS = Enum(
+    'openstack_actions_runner_status', 'Metrics displaying the status of a runner',
+    states=['creating', 'deleting', 'respawning', 'online', 'running', 'offline'],
+    labelnames=['name', 'flavor', 'image'],
+)
+
```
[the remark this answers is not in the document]
tcarmet: knowing which repo is being executed for each runner would be great, but I think it can be added on later on,

runners_manager/runner/RunnerManager.py
tcarmet: Me neither, I can add a function in the runner that can take the status from github and do this calculation inside the runner

runners_manager/metrics/Runner.py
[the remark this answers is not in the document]
tcarmet: Regarding the class, I don't know if we should yet, I feel like people tend not to. But I may be wrong.

Have a look at the documentation:
* https://prometheus.io/docs/instrumenting/writing_clientlibs/
* https://github.com/prometheus/client_python 

I feel like we should use those objects as vars inside their dedicated class. Let me know what you think

charts/openstack-actions-runner/templates/service-manager.yaml
```diff
@@ -0,0 +1,16 @@
...
+    {{- include "openstack-actions-runner.labels" . | nindent 4 }}
+spec:
+  type: {{ .Values.deployment.service.type }}
+  ports:
+    - name: metrics
+      port: {{ .Values.metricsPort }}
+      protocol: TCP
+  selector:
+    role: manager
+    {{- include "openstack-actions-runner.selectorLabels" . | nindent 4 }}
```
tcarmet, line 16: Creating a service so that the service monitor object can discover the endpoint to monitor

runners_manager/metrics/Runner.py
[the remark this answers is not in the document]
tcarmet: I'm gonna wait for your thoughts, but I think I'm gonna move this var to be inside the runner class.

runners_manager/runner/Runner.py
tcarmet: ok lol

runners_manager/runner/Runner.py
```diff
@@ -80,21 +89,40 @@ def toJson(self):
...
 
+        PROMETHEUS_RUNNER_STATUS.labels(
+            name=self.name,
+            flavor=self.vm_type.flavor,
+            image=self.vm_type.image
+        ).state(self.status)
+
+        # TODO: Find why the __delete__ function is not called or why __del__ is called all the time.
+        if self.status == 'deleting':
+            PROMETHEUS_RUNNER_STATUS.remove(self.name, self.vm_type.flavor, self.vm_type.image)
```
tcarmet, line 112: I need help figuring it out this one @erwan-b, if you have some free time in your day. Ideally I want to use the `__del__` or `__delete__` class to perform this call

runners_manager/runner/RunnerManager.py
tcarmet: I tried solving this issue, checkout the code and let me know what you think

runners_manager/metrics/Runner.py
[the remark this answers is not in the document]
erwan-b: Event prometheus, in there Python client, give class for more complex tracking: 
- `from prometheus_client.multiprocess import MultiProcessCollector`
- `from prometheus_client import CollectorRegistry`

Because as they say "Take advantage of each language’s features." When we will start to have more complex dashboard with in a way duplicate data in different format, i am certain this will help.

runners_manager/runner/Runner.py
erwan-b: I would prefer not.

As we are using a database to store the runners, the object is the a representation of the database state, the object is not the runner.
There for, the object runner can potentially be created / deleted regardless the database state.

tcarmet: Understood. I'll keep the following logic then.

runners_manager/metrics/Runner.py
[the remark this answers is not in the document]
tcarmet: Added the metrics inside a class.

runners_manager/monitoring/prometheus.py
```diff
@@ -0,0 +1,11 @@
...
+
+from prometheus_client import Enum
+
+
+class Metrics(object):
+    runner_status = Enum(
+        'openstack_actions_runner_status', 'Metrics displaying the status of a runner',
+        states=['creating', 'deleting', 'respawning', 'online', 'running', 'offline'],
+        labelnames=['name', 'flavor', 'image'],
+    )
```
erwan-b, line 11: This class should evolve with function when it growth

erwan-b (approved):

runners_manager/monitoring/prometheus.py
tcarmet: Indeed, I agree, especially when reusing labels and all. Once we add more metrics we should keep that in mind.

PR #50, opened 2021-11-04
Setup pytest and test coverage reporting

tcarmet:
- Change testing framework to pytest
- Add testing coverage report


erwan-b (approved):

erwan-b:
Is it normal to have so many missing data ?

tcarmet:
> Is it normal to have so many missing data ?

if you're referring to the missing data on the main branch, that's because we never uploaded a report on the main branch. So there's nothing to compare

PR #52, opened 2021-11-05
[RELENG-4539] ADd webhook management

erwan-b:


tcarmet (commented):
I understand that on one side you're making a "package" and on the other side you're making a web API that use the package.

I wonder if it's too early in the state of the runner manager to make this change and I think I would rather see them merged, for now at least. So I wonder what are the benefits you see by doing this?

Second, I don't see any kind of authentication for the webhook, have you looked into it? Here's what I know about it:
* github webhooks can be configured with a secret that can be sent by github in the http request. 
* GitHub webhook service also contain a specific set of IPs I believe. They can be retrieved like the following `curl https://api.github.com/meta | jq ".hooks"` we could whitelist those. But I rather go with option 1.

Besides those comment, everything looks pretty good, as you mentioned this morning I would like to see some kind of reconciliation loop just for the runner to ensure that everything is looking good even if we don't receive any webhook.

web/Webhook.py
```diff
@@ -0,0 +1,50 @@
...
+    event: str
+    payload: dict
+
+    def __init__(self, payload: dict, event: str):
+        self.event = event
+        self.payload = payload
+
+    def __call__(self, *args, **kwargs):
+        # Check if we managed this event
+        if self.event not in [methode for methode in dir(self) if methode[:2] != "__"]:
```
tcarmet, line 18:
```suggestion
        if self.event not in [method for method in dir(self) if method[:2] != "__"]:
```

web/Webhook.py
```diff
@@ -0,0 +1,50 @@
...
+            logger.info(f'Get event: {event}')
+            getattr(self, self.event)(self.payload)
+
+    def workflow_run(self, payload):
+        pass
+
+    def workflow_job(self, payload):
+        status = {}
+        if payload['action'] == 'queued' or "self-hosted" in payload["workflow_job"]["labels"]:
+            return
```
tcarmet, line 30: Help me out a bit here :sweat_smile: 
1. Shouldn't it be an `and` in the condition?
```suggestion
        if payload['action'] == 'queued' and "self-hosted" in payload["workflow_job"]["labels"]:
            return
```
2. Isn't this the moment where we can ask the runner manager to create an additional VM?

web/main.py
```diff
@@ -16,3 +18,26 @@ async def root():
...
+async def reset_reset_runners(request: Request):
+    runner_m.remove_all_runners()
+    g_runners = github_manager.get_runners()
+    runner_m.update_all_runners(g_runners['runners'])
+
+
+@app.post('/webhook')
+async def webhook_post(payload: str = Form(...), request: Request = None):
+    Webhook(payload=json.loads(payload), event=request.headers['X-Github-Event'])()
+    return Response(status_code=200)
```
tcarmet, line 43: OPTIONAL: if you add something like this:
```suggestion
    return Response(status_code=200)


@app.delete('/runners/<runner name>')
async def delete_runner(request: Request):
    runner_m.delete_runner(runner.name)
    return Response(status_code=202)
```

I'll buy you a beer :beers: :smile:

web/main.py
```diff
@@ -16,3 +18,26 @@ async def root():
...
+    except Exception as e:
+        logger.info(e)
+    return Response(status_code=200)
+
+
+@app.post('/runners/reset')
+async def reset_reset_runners(request: Request):
+    runner_m.remove_all_runners()
+    g_runners = github_manager.get_runners()
+    runner_m.update_all_runners(g_runners['runners'])
```
tcarmet, line 37: Should we add a `Response` here?

PR #55, opened 2021-11-08
[RELENG-4539] User webhooks to trigger the manager

erwan-b:


charts/openstack-actions-runner/templates/deployment.yaml
```diff
@@ -36,8 +35,8 @@ spec:
           {{- toYaml .Values.securityContext | nindent 12 }}
         image: "{{ .Values.image.manager.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
-        command: ["runner-manager"]
-        args: ["--settings-file", "/app/config/settings.yml"]
+        command: ["uvicorn"]
+        args: ["srcs.web.app:app", "--host", "0.0.0.0", "--port", "80"]
```
tcarmet, line 39: Why are we not making this command the default for the CMD on the Docker image?

tcarmet (commented):
Overall the logic looks good to me.

I think there's something to be addressed on consistency with the chart. Either we remove the "role: manager" or we keep it. I suggest to remove it everywhere. 

Also there was one logic regarding the update of action_id that is duplicated, I left a comment I'll let you double check that one.

PR #66, opened 2021-11-15
Create new class for Redis database management 

erwan-b:


tcarmet (commented):
I like the concept!

Especially having a redismanager class will be very useful!

I wonder if we actually need the manager key in redis, it seems to be used only to list runners, something that we can do depending on how we store the key, maybe. You tell me.

I'll re review and approve once we're out of WIP.

tcarmet (approved):

PR #71, opened 2021-11-17
Correct service monitor port with the according name

tcarmet:


erwan-b (approved):

erwan-b:
Small explanation about this ?

tcarmet:
I first thought that the service monitor would take the name of the port at the pod level, actually it's different.

There's three type of objects that are taken into account:
* Pod
* Endpoint
* Service

The service object creates endpoints that  contain the same value as the service but one endpoint is attached to one pod, but one service is attached to multiple pod.
The service monitor use the endpoint object to retrieve the pod info (ip and port) and as this endpoint object contain the port name from the service level I need to use the same port name as the service on the port object.

PR #74, opened 2021-11-24
[RELENG-5159] Leaking virtual machines and small improvments

erwan-b:


srcs/runners_manager/runner/Manager.py
```diff
@@ -39,7 +39,8 @@ def __init__(self, settings: dict,
     def remove_all_runners(self):
         for manager in self.runner_managers:
             for runner in manager.get_runners().values():
-                manager.delete_runner(runner)
+                if not runner.is_running:
+                    manager.delete_runner(runner)
```
tcarmet, line 43: so now when we reset, we don't delete runners that have a job ongoing?

srcs/runners_manager/runner/RunnerFactory.py
```diff
@@ -96,7 +97,13 @@ def delete_runner(self, runner: Runner):
             logger.info(f'APIException catch, when try to delete the runner: {str(runner)}')
 
     def generate_runner_name(self, vm_type: VmType):
+        from web import redis_database
```
tcarmet, line 100: I don't know if I like import at the function level, can we move this up?

srcs/runners_manager/runner/RunnerFactory.py
```diff
@@ -96,7 +97,13 @@ def delete_runner(self, runner: Runner):
...
         vm_type.tags.sort()
-        return self.runner_name_format.format(index=self.runner_counter,
+        name = self.runner_name_format.format(index=self.runner_counter,
                                               organization=self.github_organization,
                                               tags='-'.join(vm_type.tags))
+        self.runner_counter += 1
+        # Check that a virtual machine hasn't this name already
+        if redis_database.redis.get(f'runners:{name}'):
+            return self.generate_runner_name(vm_type)
+        return name
```
tcarmet, line 109: I don't get it here, it feels like we're gonna loop forever, what am I missing? Coffee :coffee: ? :smile:

srcs/runners_manager/vm_creation/openstack.py
```diff
@@ -68,38 +68,63 @@ def create_vm(self, runner: Runner, runner_token: int or None,
...
-            time.sleep(2)
-        if instance.status == 'ERROR':
-            logger.info('vm failed, creating a new one')
-            self.delete_vm(instance.id)
-            time.sleep(2)
+        # Delete all VMs with the same name
+        vm_list = self.nova_client.servers.list(search_opts={'name': runner.name},
+                                                sort_keys=['created_at'])
+        for vm_id in vm_list:
+            self.nova_client.servers.delete(vm_id)
```
tcarmet, line 81: I find it a little bit surprising that nova_client.servers.list returns a list of id, I would guess that we would have a Object list and that this loop would look more like this:
```suggestion
        vm_list = self.nova_client.servers.list(search_opts={'name': runner.name},
                                                sort_keys=['created_at'])
        for vm in vm_list:
            self.nova_client.servers.delete(vm['id'])
```

I'm pretty sure you know what you are doing but could you please just double check my statement here :smile:

templates/init_runner_script.sh
```diff
@@ -36,4 +36,4 @@ sudo -H -u actions bash -c 'mkdir -p /home/actions/actions-runner'
 sudo -H -u actions bash -c 'cd /home/actions/actions-runner && curl -O -L {{ installer["download_url"] }} && tar xzf ./{{ installer["filename"] }}'
 sudo -H -u actions bash -c 'sudo /home/actions/actions-runner/bin/installdependencies.sh'
 sudo -H -u actions bash -c '/home/actions/actions-runner/config.sh --url https://github.com/{{ github_organization }} --token {{ token }} --name "{{ name }}" --work _work  --labels {{ tags }} --runnergroup {{ group }} --replace --unattended'
-nohup sudo -H -u actions bash -c '/home/actions/actions-runner/run.sh --once 2> /home/actions/actions-runner/logs'
+nohup sudo -H -u actions bash -c '/home/actions/actions-runner/run.sh --once 2> /home/actions/actions-runner/logs'
```
tcarmet, line 39:
```suggestion
nohup sudo -H -u actions bash -c '/home/actions/actions-runner/run.sh --once 2> /home/actions/actions-runner/logs'

```

srcs/runners_manager/vm_creation/openstack.py
```diff
@@ -68,38 +68,63 @@ def create_vm(self, runner: Runner, runner_token: int or None,
...
+                userdata=self.script_init_runner(runner, runner_token, github_organization,
+                                                 installer)
+            )
+
+            while instance.status not in ['ACTIVE', 'ERROR']:
+                instance = self.nova_client.servers.get(instance.id)
+                time.sleep(2)
+
+            if instance.status == 'ERROR':
+                logger.info('vm failed, creating a new one')
```
tcarmet, line 105: Bonus point if you can add the instance.id here, that would be useful to share with support to track what happens to the creation of those vms id.

srcs/runners_manager/runner/Manager.py
erwan-b: By default github won't allow them to be deleted.

srcs/runners_manager/runner/RunnerFactory.py
erwan-b: This fonction only generate the name, this if just check that we are not using the same name than an existing runner set in the redis database

srcs/runners_manager/vm_creation/openstack.py
erwan-b: You must be correct, this part couldn't be testing because openstack was full

srcs/runners_manager/runner/Manager.py
tcarmet: true

erwan-b: BUT ! we can consider that if we use this function, we want to try in the case our database is not synchronized with github

srcs/runners_manager/runner/RunnerFactory.py
tcarmet: right, but as we are calling it recursively, won't we always hit this condition and never return the name of the runner?

erwan-b: :coffee:   time ;) 

the if condition check if it's setup in redis, if it's the case we make a recursive call. line 108 
If the redis return None, the name is not set in the database so we can return it.  line 109

tcarmet: I would see something like this:
```suggestion
        self.runner_counter += 1
        # Check that a virtual machine hasn't this name already
        runner =  redis_database.redis.get(f'runners:{name}')
        if runner:
            return runner['name']
        return name
```

erwan-b: This function genreate a name therefore it must be an un used one.
WIth this we can have multiple VM having the same name

tcarmet: aaaaaaaaaaaaaah now I see, I'm sorry. All good here

erwan-b: I will add a explanation in the doc then

srcs/runners_manager/runner/Manager.py
tcarmet: Yes, at first, for this scenario I was actually gonna suggest having a soft reset and a hard reset, but if we can't delete a runner that is running the hard reset won't work.

So all good here, this function will actually be safe to use at any time so that's good.

tcarmet (approved):
LGTM

PR #85, opened 2022-02-11
[RELENG-5430] Patch exporter data not allways removing runner

erwan-b:


tcarmet (approved):
aah ok I see. LGTM

Starting to wonder if those status needs to be called more implicitly in another way to not be forgotten. But that's for another day I would say.

PR #88, opened 2022-02-18
[RELENG-5488] Runner manager stuck in "creation" loop

erwan-b:


srcs/runners_manager/vm_creation/github_actions_api.py
```diff
@@ -26,9 +26,19 @@ def link_download_runner(self, archi='x64'):
...
+
+    def get_runners(self):
+        index = 1
+        page = self._get_runner_page(page=index, per_page=100)
+
+        while index * 100 < page['total_count']:
+            index += 1
+            page['runners'] += self._get_runner_page(page=index, per_page=100)['runners']
+
+        return page
```
bertranddemiddelaer, line 41: You are using 3 times the hardcoded value 100. It could be defined in a var at the beginning of your function.

bertranddemiddelaer (changes requested):

bertranddemiddelaer (approved):

PR #90, opened 2022-02-21
[RELENG-5484] bugfix vm leaks

erwan-b:
Create a  function that run every hours and delete vm not tracked.

### DIFF
--- a/srcs/web/app.py
+++ b/srcs/web/app.py
@@ -4,16 +4,41 @@ from fastapi import FastAPI, Request
 from fastapi.responses import Response
 from fastapi_utils.tasks import repeat_every
 
-from web import redis_database, runner_m, github_manager
+from web import redis_database, runner_m, github_manager, openstack_manager
 from web.WebhookManager import WebHookManager
 from web.models import WebHook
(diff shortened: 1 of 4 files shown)


bertranddemiddelaer (approved):

srcs/web/app.py
```diff
@@ -16,6 +16,33 @@
 app.add_route('/metrics', prometheus_metrics)
 
 
+@app.on_event("startup")
+@repeat_every(seconds=60 * 60)
+def delete_orphan_runners():
+    metrics.runner_vm_orphan_delete.set(0)
+    metrics.runner_github_orphan_delete.set(0)
```
tcarmet, line 23: If you set the metrics at 0 everytime we can this function, prometheus will not be able to track the amount of orphan you deleted.

tcarmet (changes requested):

srcs/web/app.py
erwan-b: It will set the metrics every hour. 
Because if we don't reset it, it will always increment and never stop

PR #97, opened 2022-03-28
Manage no docker tag

erwan-b:
if `no-docker` is in a tag or is a tag of the runner, we don't installer docker


bertranddemiddelaer (approved):
LGTM

PR #100, opened 2022-03-31
Add on demande vms

erwan-b:


erwan-b:
Do we need to add the max on the runner_creation  ? It could simplify some part of the code.
But it can add unwanted limit on the `on_demand` vms

srcs/runners_manager/runner/RunnerManager.py
```diff
@@ -70,6 +70,9 @@ def update_runners(self, github_runners: list[dict]) -> None:
             self.update_runner(elem)
 
     def create_runner(self) -> None:
+        if 0 < self.vm_type.quantity["max"] <= len(self.runners):
```
tcarmet, line 73: I don't know if that's a trick I'm not familiar with :D. What is this magic `0 <` ?
```suggestion
        if self.vm_type.quantity["max"] <= len(self.runners):
```

srcs/web/WebhookManager.py
```diff
@@ -49,6 +49,29 @@ def workflow_job(self, payload: WebHook):
...
+                    iter(
+                        runner_m.get_runner_manager_on_demand(
+                            lambda elem: elem.vm_type.tags
+                            == payload.workflow_job.labels
+                        )
+                    ),
+                    None,
+                )
+            except Exception as e:
+                logger.info(e)
```
tcarmet, line 64: Should it be an error here or nah?
```suggestion
                logger.error(e)
```

tcarmet (commented):
Open question is it necessary to have the field `on_demand`? shouldn't it  be true automatically if `min: 0`?

Also what happens if we have one pool for on demand on each cloud, openstack and gcloud and both receive the webhook? I think they will both tackle the creation of the runner and then github will select the one that it likes. Which we could say, we only create on demand on gcloud, that would be fine.

And lastly, I remember talking with you in the case we miss a webhook (or github forgets to send one). What are we doing here to ensure we don't have a job stuck in queue? (It's fine if we do nothing, I'm just curious to see if we have implicitly implemented something).

srcs/runners_manager/runner/RunnerManager.py
erwan-b: Because if max is equal to zero we  don't have a maximum

srcs/web/WebhookManager.py
```diff
@@ -46,9 +46,32 @@ def workflow_job(self, payload: WebHook):
...
+                logger.error(e.with_traceback())
+                return 0
+
+            logger.info(f"{r_m} Runner manager found")
+            if r_m:
+                logger.info("Start create new runner " + r_m.redis_key_name())
+                try:
+                    r_m.create_runner()
+                except Exception as e:
+                    print(e)
```
tcarmet, line 73: logger.error

tcarmet (approved):

PR #112, opened 2022-05-11
[RELENG-5952] Ensure two instance can run on the same time with diffe…

erwan-b:
…rent providers


Dockerfile
```diff
@@ -17,11 +17,8 @@ WORKDIR /app
 
 COPY poetry.lock pyproject.toml /app/
 
-RUN poetry config virtualenvs.create false && \
-    poetry install --no-interaction --no-ansi --no-dev --no-root
+RUN poetry install --no-interaction --no-ansi --no-dev --no-root
```
tcarmet, line 20: Oh we were installing twice?
Also seems like setting virtualenvs on the container, any reason why?

srcs/runners_manager/runner/RunnerFactory.py
```diff
@@ -16,24 +16,29 @@ class RunnerFactory(object):
...
+    ):
         """
         This object spawn and delete the runner and spawn the VM
         """
         self.cloud_manager = cloud_manager
         self.github_manager = github_manager
         self.github_organization = organization
-        self.runner_name_format = 'runner-{organization}-{tags}-{index}'
+        self.runner_name_format = "-{organization}-{tags}-{index}"
+        self.runner_prefix = "runner-{cloud}"
```
tcarmet, line 41: ok that's the part that we are interested in :D

tcarmet (approved):
awesome! thanks @erwan-b

tcarmet:
I borrowed your PR to work on the gcloud [backend](https://github.com/scality/openstack-actions-runner/pull/113) and noticed that there might be a couple of tweaks to review, keeping note:
* Define proper return values and object expected on the Cloudmanager, example: instance
* VmType definition needs to change and be modular depending on the cloud, I did this badly on my draft pr, in other words, probably broke the whole code. We might need to assess.


Leaving my approval as I don't think we should block anything but I think it's worth to re review the state here and define the next step together.

srcs/runners_manager/vm_creation/openstack/OpenstackManager.py
```diff
@@ -23,17 +23,20 @@ class OpenstackManager(CloudManager):
     """
     Manager related to Openstack virtual machines
     """
+    SETTING_OBJ = OpenstackConfig
```
tcarmet, line 26: Did I miss something or this needs to be renamed?
```suggestion
    CONFIG_SCHEMA = OpenstackConfig
```

Dockerfile
erwan-b: if we add this, we add an error during the installation

PR #113, opened 2022-05-12
RELENG-5962 implement gcloud backend

tcarmet:


srcs/runners_manager/vm_creation/VmType.py
```diff
@@ -4,15 +4,13 @@ class VmType:
...
-    image: str
+    config: dict
     quantity: dict[str, int]
 
     def __init__(self, config):
         config["tags"].sort()
         self.tags = config["tags"]
-        self.flavor = config["flavor"]
-        self.image = config["image"]
+        self.config = config["config"]
```
erwan-b, line 13: Exactly what we where talking about. 
100% agreed with this, but we need to edit the `OpenstackManager.py` in this case

tcarmet: yeah I pretty much broke the whole code as I didn't adapt the changes at all.

config_example.yml
```diff
@@ -4,14 +4,14 @@ github_organization: ""
...
-    token: ""
-    network_name: ""
+  auth_url: ""
+  region_name: ""
+  project_name: ""
+
+  username: ""
+  password: ""
+  token: ""
+  network_name: ""
```
erwan-b, line 14: This is Openstack cloud config, we should delete it and link to the cloud folder :thinking:

erwan-b:
+ I think we should find a solution too keep the tags in the runners names

PR #173, opened 2023-02-14
💚 Fix Unit Test

gaspardmoindrot:


tcarmet (approved):
Uhul 🎉 Perfect, thank you for this!

If you can add some context in the PR description explaining why we now need two arguments that would be great 👌

gaspardmoindrot:
We had the error `TypeError: Manager.need_new_runner() missing 1 required positional argument: 'manager'`, on closer inspection we realized that `need_new_runner()` needed 2 arguments, the first one referring to `self` and the second one referring to `manager`.

To solve this error, we have to call the method `need_new_runner()` with `r` - the instance of manager - and `r` itself.

PR #175, opened 2023-02-15
💚 Fix Unit test V2

gaspardmoindrot:


tcarmet (approved):
thank you for this, makes more sense now :)

PR #184, opened 2023-02-21
🐛 Adding optional awsSessionToken to secre.yaml for AWS runner-manager

gaspardmoindrot:


tcarmet (changes requested):
Careful with the syntax, try to play with `helm template` to ensure it looks good with the value set and unset.

Missing also the same mechanism for:
* REDHAT_USERNAME
* REDHAT_PASSWORD

According to the error we got:
```shell
error validating data: [unknown object type "nil" in Secret.data.AWS_SESSION_TOKEN, unknown object type "nil" in Secret.data.REDHAT_PASSWORD, unknown object type "nil" in Secret.data.REDHAT_USERNAME]
```

tcarmet (approved):

PR #191, opened 2023-02-23
 runner-manager values file for gcloud

Abubakarr99:


charts/runner-manager/values-runner-manager-gcloud.yaml
```diff
@@ -0,0 +1,57 @@
...
+githubToken: "vault:secret/data/runner-manager#GITHUB_TOKEN"
+redhatUsername: "vault:secret/data/runner-manager#REDHAT_USERNAME"
+redhatPassword: "vault:secret/data/runner-manager#REDHAT_PASSWORD"
+#redis:
+#  password: "vault:secret/data/runner-manager#rREDIS_PASSWORD"
+#allowedSshKeys: "vault:secret/data/runner-manager#allowed_ssh_keys"
+githubOrganization: scalanga-devl
+deployment:
+  ingress:
+    #enabled: true
```
gaspardmoindrot, line 11: Why did you comment on the whole igress? If it's done on purpose to test the service without and then implement it, I'm fine with it

charts/runner-manager/values-runner-manager-gcloud.yaml
```diff
@@ -0,0 +1,57 @@
...
+  minutes: 20
+  hours: 0
+pythonConfigModule: 'srcs.settings.settings_local'
+serviceMonitor:
+  enabled: true
+  labels:
+    monitoring-instance: prometheus-central
+
+image:
+  tag: '0.4.0'
```
gaspardmoindrot, line 34: You should make another release if you push new files

charts/runner-manager/values-runner-manager-gcloud.yaml
Abubakarr99: yes exactly to test the service

charts/runner-manager/values-runner-manager-gcloud.yaml
Abubakarr99: Ok

Abubakarr99:
It seems we won't do this, so I am closing the PR

PR #193, opened 2023-02-23
⬆️  Adding static analysis with CodeQL & Dependabot & Trivy

gaspardmoindrot:


.github/workflows/codeQL-analysis.yaml
```diff
@@ -0,0 +1,24 @@
+---
+name: Static analysis with CodeQL
```
tcarmet, line 2: Let's keep name shorts

* rename file to `codeql.yaml`
and: 
```suggestion
name: codeql
```

.github/workflows/trivy-scan.yaml
```diff
@@ -0,0 +1,23 @@
...
+    name: Trivy Vulnerability Scan
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Aqua Security Trivy Scan
+        uses: aquasecurity/trivy-action@0.9.1
+        with:
+          image-ref: python:3.9.13-slim
+          format: table # more readable for humans
```
tcarmet, line 23: We want to analyse `ghcr.io/scality/runner-manager:${{ github.sha }}` this image is not enough.

.github/workflows/codeQL-analysis.yaml
```diff
@@ -0,0 +1,24 @@
...
+  analyze:
+    name: Static analysis with CodeQL
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v2
+        with:
+          languages: python
```
tcarmet, line 22: Isn't there any more language we may find interesting?

.github/dependabot.yml
```diff
@@ -5,8 +5,35 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    commit-message:
+      prefix: "pip prod"
+      prefix-development: "pip dev"
+      include: "scope"
+    labels:
+      - "pip"
+      - "dependencies"
```
tcarmet, line 14: Let's dig more into this setting.

.github/workflows/trivy-scan.yaml
tcarmet, line 23:
```suggestion
          format: table # more readable for humans

```

tcarmet, line 23:
```yaml
      - name: Run Trivy vulnerability scanner
        uses: aquasecurity/trivy-action@7b7aa264d83dc58691451798b4d117d53d21edfe
        with:
          image-ref: 'docker.io/my-organization/my-app:${{ github.sha }}'
          format: 'template'
          template: '@/contrib/sarif.tpl'
          output: 'trivy-results.sarif'
          severity: 'CRITICAL,HIGH'
      - name: Upload Trivy scan results to GitHub Security tab
        uses: github/codeql-action/upload-sarif@v2
        with:
          sarif_file: 'trivy-results.sarif'

```

.github/workflows/codeQL-analysis.yaml
gaspardmoindrot: Done

.github/workflows/trivy-scan.yaml
gaspardmoindrot: It's changed (in the file `build.yaml`)

.github/workflows/codeQL-analysis.yaml
gaspardmoindrot: I don't think there are any other languages in the 'runner-manager' case

.github/dependabot.yml
gaspardmoindrot:
```
branches:
      - "dev*"
```
To run Dependabot on other branches

.github/workflows/trivy-scan.yaml
gaspardmoindrot: I did it differently, check and tell me if it's ok for you, if not I'll change it according to your comments!

.github/workflows/build.yaml
```diff
@@ -35,3 +38,28 @@ jobs:
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Aqua Security Trivy Scan
+        uses: aquasecurity/trivy-action@0.9.1
+        id: trivy_scan
+        with:
+          image-ref: ghcr.io/${{ github.repository }}:${{ github.sha }}, redis:alpine
+          format: table # more readable for humans
```
tcarmet, line 55: I don't think codeql will accept this format, it has to be sarif.

gaspardmoindrot: Oh ok, i will change this

gaspardmoindrot: I changed to json, normally it will work now

gaspardmoindrot:
There is an error with the trivy build, I will check this tomorrow morning

.github/workflows/build.yaml
```diff
@@ -35,3 +38,58 @@ jobs:
           tags: "ghcr.io/${{ github.repository }}:${{ github.sha }}"
           cache-from: type=gha,scope=${{ github.repository }}
           cache-to: type=gha,scope=${{ github.repository }},mode=max
+
+  trivy-scan-code:
```
tcarmet, line 42:
```suggestion
  trivy:
```

.github/workflows/build.yaml
```diff
@@ -35,3 +38,58 @@ jobs:
...
+        uses: github/codeql-action/upload-sarif@v2
+        with:
+          sarif_file: 'trivy-image-results-ghcr.sarif'
+          category: 'image-ghcr'
+      
+      - name: Upload Trivy Image Redis scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@v2
+        with:
+          sarif_file: 'trivy-image-results-redis.sarif'
+          category: 'image-redis'
```
tcarmet, line 95: Can you try translating this job into a reusable workflow? It will be useful for the tasks that will come next.

Here's the [doc](https://docs.github.com/en/actions/using-workflows/reusing-workflows#example-reusable-workflow)

You can create a new file called `.github/workflows/trivy.yaml` and call it with:
`uses: ./.github/workflows/trivy.yaml`

.github/workflows/build.yaml
```diff
@@ -35,3 +38,58 @@ jobs:
...
+
+      - name: Image Scan redis
+        uses: aquasecurity/trivy-action@master
+        with:
+          image-ref: 'redis:alpine'
+          format: 'sarif'
+          output: 'trivy-image-results-redis.sarif'
+          exit-code: '1'
+          ignore-unfixed: true
+          vuln-type: 'os,library'
```
tcarmet, line 83: While it is interesting and relevant to add the scan of this image. And we should be taking it into account I would like us to take a deeper reflection on external images/components. 

For now I suggest we remove the scan of redis, and address this topic later on:
```suggestion
```

.github/workflows/build.yaml
```diff
@@ -35,3 +38,58 @@ jobs:
...
+          category: 'code'
+
+      - name: Image Scan ghcr
+        uses: aquasecurity/trivy-action@master
+        with:
+          image-ref: 'ghcr.io/${{ github.repository }}:${{ github.sha }}'
+          format: 'sarif'
+          output: 'trivy-image-results-ghcr.sarif'
+          exit-code: '1'
+          ignore-unfixed: true
```
tcarmet, line 72: Can you remind me what `ignore-unfixed` do? And why are we using it?

.github/workflows/build.yaml
```diff
@@ -35,3 +38,58 @@ jobs:
...
+          sarif_file: 'trivy-results.sarif'
+          category: 'code'
+
+      - name: Image Scan ghcr
+        uses: aquasecurity/trivy-action@master
+        with:
+          image-ref: 'ghcr.io/${{ github.repository }}:${{ github.sha }}'
+          format: 'sarif'
+          output: 'trivy-image-results-ghcr.sarif'
+          exit-code: '1'
```
tcarmet, line 71: When does trivy exit with one? and why should we use it?

.github/workflows/build.yaml
```diff
@@ -35,3 +38,58 @@ jobs:
...
+          output: 'trivy-image-results-redis.sarif'
+          exit-code: '1'
+          ignore-unfixed: true
+          vuln-type: 'os,library'
+
+      - name: Upload Trivy Image Ghcr scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@v2
+        with:
+          sarif_file: 'trivy-image-results-ghcr.sarif'
+          category: 'image-ghcr'
```
tcarmet, line 89: What do you consider as a category? The fact that it's a image or where it is located?

ghcr is a registry but it could be published elsewhere, should we change the category if we decide to change where we publish the image?

.github/workflows/build.yaml
```diff
@@ -35,3 +38,58 @@ jobs:
...
+          security-checks: 'vuln,secret,config'
+          ignore-unfixed: true
+          format: 'sarif'
+          output: 'trivy-results.sarif'
+
+      - name: Upload Trivy scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@v2
+        with:
+          sarif_file: 'trivy-results.sarif'
+          category: 'code'
```
tcarmet, line 63: What are you scanning here? and why?

.github/workflows/build.yaml
tcarmet, line 95: Due to my comment above regarding the redis image:
```suggestion
```

tcarmet (commented):
The dependabot and codeql part LGTM.

The trivy workflow is on a good track, but need some changes.

I would advise you to split this PR in two or three, so that you can merge the dependabot and codeql change while we continue working on the changes regarding trivy.

.github/workflows/build.yaml
gaspardmoindrot: This flag excludes vulnerabilities for which there are no fixed versions. The idea is that if you can't do anything, you might as well not make a mistake

.github/workflows/build.yaml
gaspardmoindrot: With the flag exit-code: '1' if there is an error, the build with turn red on the PR.

.github/workflows/build.yaml
gaspardmoindrot: It's used for metadata and reporting purposes, you can write whatever you want in it, it will not change anything

.github/workflows/build.yaml
gaspardmoindrot: We are scanning our code with Trivy : "This includes vulnerability, configuration (as in IaC, docker files, etc), and secret scanning". If you want further info you can check this documentation https://www.pinewise.com/from-zero-to-security-scanning-in-15-minutes-with-github-actions-and-trivy/. It was an extra thing that I thought was nice to add

.github/workflows/build.yaml
gaspardmoindrot: I'm changing it

.github/workflows/build.yaml
tcarmet: It is nice indeed, let's keep it 👌

.github/workflows/build.yaml
tcarmet: 1. Double check the criteria, where is it supposed to fail? On this particular step? If so, shouldn't it be failing right now?

2. We already have a build failure that is set thanks to codeql. codeql have a database (for issues marked as won't fix and so on), therefore the whole context. And trivy on its own, will not be aware of this information. Therefore we should restrain this step from failing and leave this review to codeql.

The reason why we would like this step to fail, is if trivy was unable to perform the scan for whatever reason.

Let me know what you think with this logic.

.github/workflows/build.yaml
tcarmet: Shouldn't we report them anyway? And leave the user to decide if the particular alert should be dismissed or to keep an eye on.

.github/workflows/trivy.yaml
```diff
@@ -0,0 +1,50 @@
+---
+name: trivy
+
+on:
+  workflow_call:
+    inputs:
+      image-ref:
+        required: true
+        type: string
```
tcarmet, line 9: Optional, in case you would like to set a default:
```suggestion
    inputs:
      image-ref:
        default: "ghcr.io/${{ github.repository }}:${{ github.sha }}"
        required: false
        type: string
```

.github/workflows/build.yaml
tcarmet: It may not change stuff, but having extra metadata is super useful, so I'm glad you were able to find and make use of this option.

Now, I see you changed it as `image`. I want you to do a couple of research to see if `image` is the correct term. Ideally I would like us to use the `category` option everywhere and make use of the same term for each type of scans. (As we may find ourself filtering across multiple repo one day).

.github/workflows/build.yaml
gaspardmoindrot: Ok I see, I don't necessarily agree with you. The idea is that the security processes are implemented thoroughly (even if it's a bit annoying for the user) and the fact that the build is red will put an additional alert that will unconsciously force the user to modify his code at least a bit before pushing

.github/workflows/build.yaml
tcarmet: Then check my comment on https://github.com/scality/runner-manager/pull/193#discussion_r1119367159

And review back this story

.github/workflows/build.yaml
gaspardmoindrot: Ok I see! I will double check and found the perfect term. Do you have in mind something like this : `<prefix>-<name_of_répo>` ?

.github/workflows/trivy.yaml
gaspardmoindrot: Ohhh good idea. But, will it be useful for our case ?

.github/workflows/build.yaml
gaspardmoindrot: Yeah, I think we can discuss it together on Google Meet, it will be easier for taking a decision

.github/workflows/build.yaml
tcarmet: Finding the perfect term is an impossible task, 😂 so don't worry too much about it. Just some extra thought to ensure you are happy with the name.

I honestly think that `category: image` is good. Other options I can think of is `docker` or `container`. I don't have a strong opinion for now, use what you think fits. 

Don't include the name of the repo, as we want the category to be named the same across different repo, allowing to do searches like `org:scality category:image severity:critical` and get the list accordingly. 

The thing to remember here is every time you will add metadata on this security topic, try to give it some thought on how it could be used.

gaspardmoindrot: Ohh ok you're right. I think `image` is a good term, if I find something else I will change

.github/workflows/trivy.yaml
tcarmet: Depends on how you see it, default values are always good when you can have them.
But it's better to have no default value than a bad one

In this case this would remove the need of having those two lines in the `build` workflow:
```yaml
    with:
       image-ref: 'ghcr.io/${{ github.repository }}:${{ github.sha }}'
```

.github/workflows/trivy.yaml
```diff
@@ -0,0 +1,50 @@
...
+          sarif_file: 'trivy-results.sarif'
+          category: 'code'
+
+      - name: Image Scan
+        uses: aquasecurity/trivy-action@master
+        with:
+          image-ref: ${{ inputs.image-ref }} #'ghcr.io/${{ github.repository }}:${{ github.sha }}'
+          format: 'sarif'
+          output: 'trivy-image-results.sarif'
+          #exit-code: '1'
```
tcarmet, line 40:
```suggestion
```

tcarmet (approved):

gaspardmoindrot:
@nootal

PR #212, opened 2023-02-28
🐛 Adding base64 encode to GCloud Provider secrets

gaspardmoindrot:


charts/runner-manager/templates/google-service-account.yaml
```diff
@@ -5,5 +5,5 @@ metadata:
   name: {{ .Release.Name }}-gcloud-service-account
 data:
   # The base64 encoded json file which contains the key
-  credentials.json:   {{ .Values.gcloudServiceAccount | quote }}
+  credentials.json:   {{ .Values.gcloudServiceAccount | b64enc }}
```
tcarmet, line 8:
```suggestion
  credentials.json: {{ .Values.gcloudServiceAccount | b64enc }}
```

tcarmet (commented):
Let's remove the `base64` mention in the comment above, just stating that we need `The encoded json file`

Double check also the `values` file to ensure that if there's any description it's up to date.

tcarmet (approved):

gaspardmoindrot:
> Let's remove the `base64` mention in the comment above, just stating that we need `The encoded json file`
> 
> Double check also the `values` file to ensure that if there's any description it's up to date.

Everything's okay for me even in the [`values` file](https://github.com/scality/devinfra/blob/main/applications/gcloud-runner-manager.yaml)

PR #215, opened 2023-02-28
🐛 Fix code scanning alert - Only root user on Dockerfile

gaspardmoindrot:


.devcontainer/Dockerfile
```diff
@@ -13,6 +13,12 @@ RUN poetry config virtualenvs.create false
 
 WORKDIR /workspaces/runner-manager
 
+# Create a non-root user
+
```
tcarmet, line 17: In this image, it's not used in runtime, I would say we can remove any change and solve the security issue with "used in tests"

Dockerfile
```diff
@@ -24,4 +24,8 @@ RUN poetry config virtualenvs.create false && \
 
 COPY . /app/
 
+RUN useradd -r -u 1000 -g root appuserRunnerManager \
+    && chown -R appuserRunnerManager /app
+USER appuserRunnerManager
+
```
tcarmet, line 30: First let's keep the name short

```suggestion
RUN useradd -r -u 1000 -g root runner-manager \
    && chown -R runner-manager /app
USER runner-manager

```

Second the important thing of setting a user is to remove any sort of root access. I'm not 100% sure on what the `-g root` option those, but if it adds root access to the image it should be removed.

And if you need root access review why, might be because you are using a 80 port, you can change it to 8080 to sort this out.

.devcontainer/Dockerfile
gaspardmoindrot: Oh ok, I see, you're right, we can do this

gaspardmoindrot: Done : https://github.com/scality/runner-manager/security/code-scanning/72

Dockerfile
gaspardmoindrot: You're right, I will change the `-g root` option. I will try without changing the port first, then I will see if it works or no

tcarmet (approved):

PR #219, opened 2023-03-01
🐳 Fix Dockerfile HEALTHCHECK

gaspardmoindrot:


Dockerfile
```diff
@@ -28,4 +28,6 @@ RUN useradd -r -u 1000 runner-manager \
     && chown -R runner-manager /app
 USER runner-manager
 
+HEALTHCHECK --interval=30s CMD curl --fail http://localhost:80/ || exit 1
+
```
tcarmet, line 32: I would have dismissed this alert because we do healthchecks on the k8s deployment, and they happen at the pod level.

tcarmet (approved):
LGTM

Dockerfile
tcarmet: You can leave it, I believe it will be ignored when running on kubernetes, and will be used if you use plain docker.

What I don't like about docker's healthcheck is that, per example in your case you have to install `curl` on the container, which is an extra thing you'll have to worry about now security wise. (btw is it installed?)

gaspardmoindrot: Yes it is :
```
RUN apt-get update && \
    apt-get install --no-install-recommends --assume-yes \
    ca-certificates \
    curl \
    git
```
It was already installed, I thought it was not a big deal to add the Healthcheck tho. I will close the PR

tcarmet: interesting, I wonder why we need curl here.

Anyway, up to you really, there's no harm in merging this PR. Especially as we already have curl inside.

PR #222, opened 2023-03-01
📍 Switching to 17.8.2 redis version

gaspardmoindrot:


tcarmet (commented):
* feel free to upgrade helm to the latest version in the action.
* Remove the redis chart tgz as a whole, as discussed, it doesn't need to be committed. (I think).
* Does it work? Have you tested it?

PR #226, opened 2023-03-13
Review dependency review execution

gaspardmoindrot:


.github/workflows/codeql.yaml
```diff
@@ -107,3 +107,6 @@ jobs:
 
       - name: 'Dependency Review'
         uses: actions/dependency-review-action@v3
+        with:
+          base_ref: ${{ github.base_ref }}
+          head_ref: ${{ github.head_ref }}
```
tcarmet, line 112: Interesting find, what made you add this code?

What are the default values otherwise?

Is `github.base_ref` available if we don't create a PR? (Ex: a workflow that is triggered on `push`?)

gaspardmoindrot: I had an error when triggering this workflow. The error specified to add these 2 elements.

Apparently, if you don't set these variables, they may be `null` or `undefined`. 

I think the explanation is that the workflow need to compare changes between the base branch and the head branch. So if one of them is `null` or `undefined`, an error will be triggered.

I checked and no, it won't result in an error. The `github.base_ref` context variable is available during both push and pull request events, and it will simply return the base branch name for a pull request event and the current branch name for a push event.

tcarmet: [GitHub](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) docs is telling me a different story:

> The `base_ref` or target branch of the pull request in a workflow run. This property is only available when the event that triggers a workflow run is either pull_request or pull_request_target.

From what I'm understanding this code, as is, will fail if it is not triggered by a `pull_request` event. You can double check by triggering a workflow_dispatch on your branch. See if it works or not.

Now I think we need to understand if we actually need to execute this action on events other than PRs. Depending on the answer we can modify the code accordingly.

gaspardmoindrot: I close this PR, the bug is fixed thanks to this one: https://github.com/scality/runner-manager/pull/230

PR #234, opened 2023-03-29
Adding AWS tags

gaspardmoindrot:


srcs/runners_manager/vm_creation/aws/schema.py
```diff
@@ -1,5 +1,7 @@
 from marshmallow import Schema, fields
 
+class AwsCloudConfig(Schema):
+    aws_tags = fields.Dict(keys=fields.Str(), values=fields.Str())
```
tcarmet, line 4: Seems a bit repetitive to mention `aws` in a config that is specific to `aws`. I wouldn't mind the config to look like this at the end:

```yaml
cloud_name: aws
cloud_config:
  tags:
    my_tag: value
```

```suggestion
    tags = fields.Dict(keys=fields.Str(), values=fields.Str())
```
If you want to differentiate `tags` of github runners and aws tags you can use `ec2_tags` that's a legitimate specification.

And please ensure the value is set as optional (most likely is already the case, just want to double check)

tcarmet (commented):
You're heading in the right direction so the code as is LGTM.

Couple of stuff missing though:

- [ ] Update the doc accordingly to mention the tags availability.
- [ ] Ensure the helm chart is compatible with the new config (I think it is already but just double check).

srcs/runners_manager/vm_creation/aws/schema.py
gaspardmoindrot: I check and change everything once I get it to work! Thanks for the comment!

gaspardmoindrot: Done, everything's good! It is optional, I just checked

srcs/runners_manager/vm_creation/aws/AwsManager.py
```diff
@@ -27,7 +29,8 @@ def __init__(
         super(AwsManager, self).__init__(
             name, settings, redhat_username, redhat_password, ssh_keys
         )
-        self.ec2 = boto3.client("ec2")
+        self.tags = settings.get("tags")
+        self.ec2 = boto3.client("ec2", region_name='us-west-2')
```
tcarmet, line 33: We were using the `AWS_DEFAULT_REGION` environment var to set this setting. What happened?

We cannot hardcode the value in the code.
```suggestion
        self.ec2 = boto3.client("ec2")
```

tcarmet (changes requested):

srcs/runners_manager/vm_creation/aws/AwsManager.py
gaspardmoindrot: Damn I forgot to remove it or to do it properly, but indeed, I don't know why but if you don't give this parameter in addition it returns an error. I'll look for a way to fix it (or make it cleaner)

gaspardmoindrot: Here is the error : `botocore.exceptions.NoRegionError: You must specify a region.`

gaspardmoindrot: Ok, it's good now

tcarmet (approved):

PR #242, opened 2023-04-25
[RELENG-7502-aws-tag] Upgrading boto3 version

gaspardmoindrot:


nootal (changes requested):
Make sure your commit message contains the reference to the ticket, otherwise this information will be lost once the branch is merged.

config_example.yml
```diff
@@ -37,7 +35,23 @@ python_config: 'settings.settings_local'
 #      tags:
 #        - centos7
 #        - small
-runner_pool: []
+runner_pool:
```
Abubakarr99, line 38: Is it recommended to hard cord this values in a public repository ? Can't the values be updated in devinfra for example the runner pool and over here we leave it empty. I'm talking about values such as the security_group_ids, subnet_id. Or maybe I'm missing something

gaspardmoindrot: This is a mistake, I did not want to push this file! I just deleted it again

Abubakarr99 (approved):
LGTM

PR #243, opened 2023-04-25
[RELENG-7502] 🔨 Hard fix runner-manager tags

gaspardmoindrot:


srcs/runners_manager/vm_creation/aws/AwsManager.py
```diff
@@ -62,15 +62,29 @@ def create_vm(
...
+                    'Key': 'owner',
+                    'Value': 'ci'
+                },
+                {
+                    'Key': 'tool',
+                    'Value': 'runner-manager'
                 }
             ]
 
-            for key, value in self.tags.items():
```
nootal, line 68: You have some suspicion the `self.tags` dict was not properly filled?
Because I'm OK with your changes, but it's really basic and I'm wondering what would prevent the original code from working fine.

nootal (approved):
Approving as it's harmless, but I'm really wondering (cf my comment) what would be the root cause of this issue.

PR #246, opened 2023-05-01
[RELENG-7526] ✨ Adding tag in conf file

gaspardmoindrot:


tcarmet (commented):
Let's handle the case where there's no tags setup. and we should be good to go.

tcarmet (approved):

PR #248, opened 2023-05-23
[PTFE-105] 🐛 Fix AWS runner-manager tags problem in the conf file

gaspardmoindrot:
Over the past few weeks, we've encountered an issue with the AWS runner-manager. Specifically, in the code of the `runner-manager` repository, we were unable to dynamically set the tags and had to hardcode them in the code. As a result, we are working to address this issue with this pull request.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,8 +24,4 @@ RUN poetry config virtualenvs.create false && \
 
 COPY . /app/
 
-RUN useradd -r -u 1000 runner-manager \
-    && chown -R runner-manager /app
-USER runner-manager
-
 CMD ["uvicorn", "srcs.web.app:app", "--host", "0.0.0.0", "--port", "80"]
--- a/srcs/runners_manager/vm_creation/aws/AwsManager.py
+++ b/srcs/runners_manager/vm_creation/aws/AwsManager.py
@@ -31,6 +31,7 @@ class AwsManager(CloudManager):
         )
         self.tags = settings.get("tags")
         self.ec2 = boto3.client("ec2")
+        logger.debug('AWS init, configured with tags: {}'.format(self.tags))
 
     def delete_existing_runner(self, runner: Runner):
         """Delete an old runner instance from AWS if it exists."""


srcs/runners_manager/vm_creation/aws/AwsManager.py
```diff
@@ -31,6 +31,7 @@ def __init__(
         )
         self.tags = settings.get("tags")
         self.ec2 = boto3.client("ec2")
+        logger.debug('AWS runner-manager init')
```
tcarmet, line 34: Take the opportunity to add a bit more detail, like, `AWS init, configured with region: x and tags: x,y,z`

Dockerfile
```diff
@@ -24,8 +24,4 @@ RUN poetry config virtualenvs.create false && \
 
 COPY . /app/
 
-RUN useradd -r -u 1000 runner-manager \
```
tcarmet, line 27: what happened?

tcarmet (changes requested):
- [ ] Update the PR title.

- [ ] Update the description with some context about the issue

tcarmet (commented):
Test this scenario:
- Keep the runner-manager user
- Modify the port from `80` to `8080`
  - Dockerfile
  - Docker compose (maybe)
  - The k8s deployment (may be in multiple locations)
  - The k8s service
  - The k8s ingress

Dockerfile
gaspardmoindrot: I'm not exactly sure, but I believe the issue stems from the fact that port 80 is only accessible by the `root` user. Therefore, I will try to change the port from 80 to 8080 everywhere

gaspardmoindrot:
> Test this scenario:
> 
> * Keep the runner-manager user
> * Modify the port from `80` to `8080`
>   
>   * Dockerfile
>   * Docker compose (maybe)
>   * The k8s deployment (may be in multiple locations)
>   * The k8s service
>   * The k8s ingress

It's not working. If you have more information about what you're seeing, I'm very eager to hear it.

tcarmet (changes requested):
There's more content that needed in this PR. I would like to focus on solving the problem.

- You have the python version?
- The port being changed?
- And the user being removed?

Which one is required to fix the issue?

gaspardmoindrot:
> There's more content that needed in this PR. I would like to focus on solving the problem.
> 
> * You have the python version?
> * The port being changed?
> * And the user being removed?
> 
> Which one is required to fix the issue?

From what you told me yesterday, you wanted to change the structure of the Dockerfile from scratch. When testing, everything is working correctly with my changes. I will revert everything and just remove the user that is causing all the issues.

tcarmet:
> > There's more content that needed in this PR. I would like to focus on solving the problem.
> > 
> > * You have the python version?
> > * The port being changed?
> > * And the user being removed?
> > 
> > Which one is required to fix the issue?
> 
> From what you told me yesterday, you wanted to change the structure of the Dockerfile from scratch. When testing, everything is working correctly with my changes. I will revert everything and just remove the user that is causing all the issues.

The request to change the structure of the Dockerfile was related to an experiment for us to understand why we needed to remove the user, as, at this moment, we don't have the full explanation. It was an opportunity for us to review the whole structure of it and identify possible issues, at the end we didn't find it. We then concluded that we were going to make this investigation at a later date.

That being said, it doesn't answer the question I initially asked, having a look at your PR I got a partial answer, so let me rephrase it: if we focus on the problem we are trying to solve here, which one of the fixes is needed?

A. The downgrade of python from 3.10 to 3.9?
B. The removal of the `runner-manager` user?

gaspardmoindrot:
> > > There's more content that needed in this PR. I would like to focus on solving the problem.
> > > 
> > > * You have the python version?
> > > * The port being changed?
> > > * And the user being removed?
> > > 
> > > Which one is required to fix the issue?
> > 
> > 
> > From what you told me yesterday, you wanted to change the structure of the Dockerfile from scratch. When testing, everything is working correctly with my changes. I will revert everything and just remove the user that is causing all the issues.
> 
> The request to change the structure of the Dockerfile was related to an experiment for us to understand why we needed to remove the user, as, at this moment, we don't have the full explanation. It was an opportunity for us to review the whole structure of it and identify possible issues, at the end we didn't find it. We then concluded that we were going to make this investigation at a later date.
> 
> That being said, it doesn't answer the question I initially asked, having a look at your PR I got a partial answer, so let me rephrase it: if we focus on the problem we are trying to solve here, which one of the fixes is needed?
> 
> A. The downgrade of python from 3.10 to 3.9? B. The removal of the `runner-manager` user?

B. The removal of the `runner-manager` user
You then want me to put back version 3.10? 

That makes more sense, I agree. However, I didn't change the version earlier in relation to `boto3`. I know there can be compatibility issues between version `3.10` and `boto3`. I haven't investigated further, but it's a problem I've already encountered, so I wanted to avoid adding another potential issue.
I tested it with version `3.10` on another branch, and everything works fine, so I can change it back.

tcarmet:
>
> B. The removal of the `runner-manager` user You then want me to put back version 3.10?
> 
> That makes more sense, I agree. However, I didn't change the version earlier in relation to `boto3`. I know there can be compatibility issues between version `3.10` and `boto3`. I haven't investigated further, but it's a problem I've already encountered, so I wanted to avoid adding another potential issue. I tested it with version `3.10` on another branch, and everything works fine, so I can change it back.

Ok so, let's focus on B then. If we see any issue with `boto3` and `python:3.10` we can always make an extra PR to fix it.

tcarmet (approved):

tcarmet (approved):

PR #252, opened 2023-06-05
[PTFE-376] ✨ Switch GCP runners to spot instances

gaspardmoindrot:


srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
```diff
@@ -11,6 +11,7 @@
 from google.cloud.compute import InstancesClient
 from google.cloud.compute import Items
 from google.cloud.compute import Metadata
+from google.cloud.compute import Scheduling
```
nootal, line 14: Alphabetical order please :)

nootal (changes requested):
Minor comment, otherwise it's good

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
```diff
@@ -107,6 +109,9 @@ def configure_instance(
             metadata=Metadata(
                 items=[Items(key="startup-script", value=startup_script)]
             ),
+            scheduling=Scheduling(
+                preemptible=True
+            ),
             advanced_machine_features=AdvancedMachineFeatures(
```
tcarmet, line 115: This should be a configuration at the pool level, located at the `config` level.

tcarmet (changes requested):
We can't hardcode the spot creation of a VM. It needs to be a configuration that is activated explicitly by the user, and should be set as `false` by default.

nootal (approved):

docs/setup/clouds/gcloud.md
```diff
@@ -69,6 +69,7 @@ runner_pool:
     project: ubuntu-os-cloud
     family: ubuntu-2004-lts
     disk_size_gb: 20
+    spot: True
```
tcarmet, line 72:
```suggestion
    spot: true
```

tcarmet (changes requested):

srcs/runners_manager/vm_creation/gcloud/schema.py
```diff
@@ -11,3 +11,4 @@ class GcloudConfigVmType(Schema):
     project = fields.Str(required=True)
     family = fields.Str(required=True)
     disk_size_gb = fields.Str(required=True)
+    spot = fields.Boolean(missing=False)
```
tcarmet, line 14: Check how defaults work with [marshmallow](https://marshmallow.readthedocs.io/en/stable/quickstart.html#specifying-defaults)
I don't think `missing` works anymore.

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
```diff
@@ -73,6 +74,16 @@ def configure_instance(
         )
         disk_size_gb = runner.vm_type.config["disk_size_gb"]
         disk_type = f"projects/{self.project_id}/zones/{self.zone}/diskTypes/pd-ssd"
+
+        automatic_restart = True
```
gaspardmoindrot, line 78: I'm not sure if VMs are currently created with `automatic_restart=true`. If that's not the case, it's better to set it to `false` in all situations.

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
```diff
@@ -73,6 +74,16 @@ def configure_instance(
...
+        provisioning_model = "STANDARD"
+        preemptible = False
+        if "preemptible" in runner.vm_type.config:
+            preemptible = bool(runner.vm_type.config["preemptible"])
+            if preemptible:
+                provisioning_model = "SPOT"
+                automatic_restart = False
```
tcarmet, line 85: Can we have all those settings in `runner.vm_type.config` with the appropriate default values?

tcarmet (changes requested):

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
tcarmet: I don't think they are, but might be interesting to play with this value for instances that are idle and scheduled with SPOT.

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
gaspardmoindrot: Yes we can but I'm not sure it's a good idea. `automatic_restart` is False when it's a `SPOT` instance, `preemptible` is True when it's a SPOT instance. In the other hand, when it's a `DEFAULT` instance, `preemptible` is False and `automatic_restart` is True. So, it's a cause-and-effect relationship. Might be bette to just fill in the information once with `preemptible`, right?

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
gaspardmoindrot: It's impossible to use `automatic_restart=true` for a SPOT instance. But in the other scenario I think it's a good idea to have it !

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
gaspardmoindrot: For me, a SPOT instance is a new version of preemptible instances. I couldn't find the exact answer regarding what would happen if we removed the `preemptible` tag, but I thought I would directly test it with the Google Cloud Console by creating a Spot VM and only specifying "Spot". So, I conducted the experiment, and once the VM was created, I went to verify, and the preemptible variable was set to "on". I believe the two are linked.

I will test in my code to enable one and not the other to see what it will result in.

gaspardmoindrot: Capture d’écran 2023-06-09 à 10 21 43

gaspardmoindrot: `[ERROR] runner_manager: 400 POST https://compute.googleapis.com/compute/v1/projects/scality-prod-ga-runners/zones/us-east1-c/instances: Invalid value for field 'resource.scheduling.provisioningModel': 'SPOT'. preemptible=false and provisioning_model=SPOT is contradicting.`

- If I set the `preemptible` tag to `False` while it is a `SPOT` instance, it returns an error that tells us we are contradicting ourselves. There is indeed a cause-and-effect relationship.

gaspardmoindrot:
I have added `instance_termination_action` with a value of `DELETE` for `SPOT` instances. I think it's a better idea since we won't need to reuse a stopped VM in any case, so it's better to delete the VM.

srcs/runners_manager/vm_creation/gcloud/schema.py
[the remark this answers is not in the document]
gaspardmoindrot: I have reverted back to using `spot` as it makes more sense to me. Especially since we are changing the variable for both preemptible and spot instances.

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
tcarmet: ok cool, thanks for checking, so it confirms we need the `preemtible` flag as well.

tcarmet (changes requested):

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
```diff
@@ -73,6 +74,18 @@ def configure_instance(
...
+
+        automatic_restart = True
+        provisioning_model = "STANDARD"
+        instance_termination_action = "STOP"
+        preemptible = False
+        if "preemptible" in runner.vm_type.config:
+            preemptible = bool(runner.vm_type.config["spot"])
+            if preemptible:
+                provisioning_model = "SPOT"
+                automatic_restart = False
```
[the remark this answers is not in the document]
gaspardmoindrot: > Then you check if the key preemptible is set on this schema, but according to the schema preemptible is not defined

For this, it was a mistake! A change all `preemptible` variable name to `spot` and did not test again the project before pushing. Nos it's corrected!

> Then if preemptible is set you retrieve spot

It's just easier this way I think. Because `preemptible` is bool and `provisioning_model = "SPOT"` is not. For me it's more for the user experience. Easier to juste have to set `spot: true` in the conf file that something else.

> Not to mention that all this logic is not required if we were following what I previously said in this [conversation](https://github.com/scality/runner-manager/pull/252/files/ecd443df8bd0c668c4f2dbb796de1d16983676d2#r1221147629)

I know, but a lot changes from that day and I don't agree with this anymore. We can discuss about it on Monday, I'm really open to discussion but I think that `preemptible` is less understandable than `spot`. Even if in the code it does more sens to use `preemptible`.

> That being said I understand that the complexity is then added upon the user configuring the pool of runners. So I'll leave the choice up to you.

I'm really down to take 5 minutes to discuss about it on Monday Standup (either 5 min before or 5 min after it)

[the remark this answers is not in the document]
tcarmet: All your points LGTM and you have addressed the issue in the logic, but I still expect the following requested change:
> Setup a default value for spot or all variables that are non required using the [previously shared method](https://github.com/scality/runner-manager/pull/252/files/ecd443df8bd0c668c4f2dbb796de1d16983676d2#r1221169524). This will allow the code to stop checking if a key was defined.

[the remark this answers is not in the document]
gaspardmoindrot: Done! Everything's changed. Can we discuss this tomorrow? I'm not convinced by all this approach

tcarmet (changes requested):

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
```diff
@@ -73,6 +74,18 @@ def configure_instance(
...
+        provisioning_model = "STANDARD"
+        instance_termination_action = "DEFAULT"
+        preemptible = False
+        if "spot" in runner.vm_type.config:
+            preemptible = bool(runner.vm_type.config["spot"])
+            if preemptible:
+                provisioning_model = "SPOT"
+                automatic_restart = False
+                instance_termination_action = "DELETE"
+
```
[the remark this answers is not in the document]
tcarmet: Round 2 after reviewing the schema and default settings:

```suggestion
        
        preemptible = bool(runner.vm_type.config.get("spot", False))
        if preemptible:
            provisioning_model = "SPOT"
            automatic_restart = False
            instance_termination_action = "DELETE"

```

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
```diff
@@ -73,6 +74,18 @@ def configure_instance(
...
+        automatic_restart = True
+        provisioning_model = "STANDARD"
+        instance_termination_action = "DEFAULT"
+        preemptible = False
+        if "spot" in runner.vm_type.config:
+            preemptible = bool(runner.vm_type.config["spot"])
+            if preemptible:
+                provisioning_model = "SPOT"
+                automatic_restart = False
+                instance_termination_action = "DELETE"
```
tcarmet, line 87: Considering marshmallow defaults are not applied:

Option 1
```suggestion
        preemptible = bool(runner.vm_type.config.get("spot", False))
        if preemptible:
            provisioning_model = "SPOT"
            automatic_restart = False
            instance_termination_action = "DELETE"
```

docs/setup/clouds/gcloud.md
```diff
@@ -82,6 +83,10 @@ runner_pool:
     * Be aware that you need two parameter, the project and the family
 * List of [machine types] can be found on Google Cloud documentation
 * disk size can be whatever you chose to be usefull (be frugal but generous ;) )
+* The spot key in the runner pool configuration is used to enable or disable the
```
tcarmet, line 86:
```suggestion
* The `spot` key in the runner pool configuration is used to enable or disable the
```

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
gaspardmoindrot: Done!

tcarmet (approved):

PR #255, opened 2023-06-09
[PTFE-324] ✨ Update GCP VM instance tags

gaspardmoindrot:


srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
```diff
@@ -73,6 +105,13 @@ def configure_instance(
         )
         disk_size_gb = runner.vm_type.config["disk_size_gb"]
         disk_type = f"projects/{self.project_id}/zones/{self.zone}/diskTypes/pd-ssd"
+        labels = {}
+        for tag in runner.vm_type.tags:
+            if "small" in tag or "medium" in tag or "large" in tag:
+                labels["flavor"] = tag
+                break
+        labels["image"] = runner.vm_type.config["machine_type"]
```
tcarmet, line 113: The machine_type is the flavor, not the image.

Let's use the variable available, defined by the user to build the image name:
`{runner.vm_type.config["project"]}-{runner.vm_type.config["family"]}`

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
```diff
@@ -73,6 +105,13 @@ def configure_instance(
         )
         disk_size_gb = runner.vm_type.config["disk_size_gb"]
         disk_type = f"projects/{self.project_id}/zones/{self.zone}/diskTypes/pd-ssd"
+        labels = {}
+        for tag in runner.vm_type.tags:
+            if "small" in tag or "medium" in tag or "large" in tag:
+                labels["flavor"] = tag
+                break
```
tcarmet, line 112: I understand why you did this, but there's too much risk of it not working or needing to be adapted as our flavor names evolve.

let's make it easier for now and remove this need from the requirement, you can:
- Remove the `flavor` label
- Or/and set the label `machine_type` instead, which should provide the same information

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
```diff
@@ -73,6 +105,13 @@ def configure_instance(
...
+        labels["image"] = runner.vm_type.config["machine_type"]
+        labels["status"] = "offline"
```
tcarmet, line 114: Is there a `runner.status` var you can use?

srcs/web/WebhookManager.py
```diff
@@ -91,6 +91,17 @@ def workflow_job(self, payload: WebHook):
                 "labels": payload.workflow_job.labels,
             }
         )
+
+        runner_m.factory.cloud_manager.add_labels_to_instance(
+            payload.workflow_job.runner_name,
+            dict(
+                status=status["status"],
+                repository=payload.repository.name,
+                workflow=payload.workflow_job.workflow_name,
```
tcarmet, line 100: Wait a minute, we now have the workflow name in the workflow job webhook? That's some good news, something we can use in the exporter to calculate costs of workflows instead of jobs.

srcs/runners_manager/vm_creation/CloudManager.py
```diff
@@ -70,6 +70,10 @@ def delete_vm(self, runner: Runner):
     def delete_images_from_shelved(self, name):
         raise NotImplementedError
 
+    @abc.abstractmethod
+    def add_labels_to_instance(self, instance_name: str, labels_webhook: dict):
```
tcarmet, line 74: What happens to the other class that don't have this method implemented? might need to add a dummy function

tcarmet, line 74: as you got confused with tags and labels because each cloud has its own way of naming things, I suggest you to review the name of this function with this in mind.

tcarmet (changes requested):

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
gaspardmoindrot: Done!

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
gaspardmoindrot: Done

srcs/web/WebhookManager.py
gaspardmoindrot: Yeahhhh!!! A ticket has been create for this for the next sprint

srcs/runners_manager/vm_creation/CloudManager.py
gaspardmoindrot: Done ! I used the word `identifier`

gaspardmoindrot: I tried something! Tell me if it's better

tcarmet: I would go for either `update_vm` or `update_vm_metadata`, here's why:
* I like the word instance better, but the rest of the code use `vm` (on the function names, the variables use instance, it's bit of a mix and match I see 😅)
* Identifiers is accurate but I don't see it used very often, metadata is more common at least for me.
* Technically we're adding some labels but we're updating others

srcs/runners_manager/vm_creation/CloudManager.py
```diff
@@ -70,6 +70,10 @@ def delete_vm(self, runner: Runner):
     def delete_images_from_shelved(self, name):
         raise NotImplementedError
 
+    @abc.abstractmethod
+    def add_identifier_to_instance(self, instance_name: str, labels_webhook: dict):
+        raise NotImplementedError
+
```
tcarmet, line 76: We don't need to state `webhook` in the name of this variable, right?

Following the logic on the other conversation I would go with `metadata` but just `labels` is fine as well.
```suggestion
    @abc.abstractmethod
    def add_identifier_to_instance(self, instance_name: str, metadata: dict):
        raise NotImplementedError

```

tcarmet (changes requested):
Only naming stuff to figure out but the rest LGTM, should be good to go in a few.

srcs/runners_manager/vm_creation/openstack/OpenstackManager.py
```diff
@@ -247,3 +247,6 @@ def delete_images_from_shelved(self, name: str):
         for i in images:
             if name in i.name:
                 self.glance.images.delete(i.id)
+
+    def update_vm_metadata(self, instance_name: str, labels_webhook: dict):
```
tcarmet, line 251:
```suggestion
    def update_vm_metadata(self, instance_name: str, metadata: dict):
```

tcarmet (changes requested):
one last change

tcarmet (approved):

PR #259, opened 2023-06-14
PTFE-421: add option to specify zone at pool level

Abubakarr99:


srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
```diff
@@ -98,6 +98,7 @@ def configure_instance(
             project=runner.vm_type.config["project"],
             family=runner.vm_type.config["family"],
         )
+        self.zone = runner.vm_type.config.get("zone", self.zone)
```
tcarmet, line 101: I don't know if we should overwrite the self here. There might be some unwanted impact. Could you have a look into that?

We also need to take the whole code into account, per example I see on the `update_vm_metadata` function we use `self.zone` to get the instances, chances are that `self.zone` and the actual zone of the instance are not the same, and the instances.get method will return that the instance does not exist (because it doesn't in that zone).

Same story for get_all_vms method.

If zone is a optional argument in those cases where we are retrieving an instance by name or listing them, then it will help us solve the issue.

srcs/runners_manager/vm_creation/gcloud/schema.py
```diff
@@ -12,3 +12,4 @@ class GcloudConfigVmType(Schema):
     family = fields.Str(required=True)
     disk_size_gb = fields.Str(required=True)
     spot = fields.Boolean(load_default=False)
+    zone = fields.Boolean(load_default=False)
```
tcarmet, line 15: Zone is not a boolean but a string.
```suggestion
    zone = fields.Str()
```

tcarmet (changes requested):
The code looks good to me, but there are some check missing to really take zone into account. (See dedicated comment on the matter).

It might be harder than we think to add zone into the picture as a parameter for the pool, if that's the case, depending on the difficulty of it, we might halt the development of the feature and make a full switch of zones. So be mindful of that when digging further.

And lastly check if the chart need some update (I don't think so but worth double checking).

gaspardmoindrot (changes requested):
Thomas has provided really valuable feedback, and I agree with him on all points! I want to add a clarification so that you don't miss the information. In addition to update_vm_metadata and get_all_vms, you also need to make changes to the delete_vm function. configure_instance is only used in create_vm, so if the runner manager doesn't need to create new instances but needs to perform a check or delete an instance, it may be in the wrong zone potentially.

Abubakarr99:
thanks for your comments I would be testing for that.  I think we might end up doing a full migration to EU and maybe later as I continue to work on adding this feature in the coming week.

Abubakarr99:
- [ ]  Check chart for config of zone/ I checked the chart the cloudconfig parameter is empty in the chart so we are good. 
- [ ] Update other methods.

Abubakarr99:
We can update some of the methods. The hard part is updating the `get_all_vms` method which is very zone specific. The method is used by the `delete_existing_runner` method. We can pass in a zone as a parameter but where is this parameter coming from. It can't be from a runner because we are trying to list all runners. We could have a list of zones supported I guess.

tcarmet:
> We can update some of the methods. The hard part is updating the `get_all_vms` method which is very zone specific. The method is used by the `delete_existing_runner` method. We can pass in a zone as a parameter but where is this parameter coming from. It can't be from a runner because we are trying to list all runners. We could have a list of zones supported I guess.
> 
> I also found a method in the instancclient class of gcp comoute that can list vms by zone.

Had a look at the [source code](https://github.com/googleapis/python-appengine-admin/blob/main/google/cloud/appengine_admin_v1/services/instances/client.py#L452)

And it doesn't seem that zone might be a parameter required, they are pretty vague about it with the mention:
> It may require specifying regional endpoints when creating the service

So test, you can create a separate small python script and run it to see how it behaves
Play a bit with it, and we'll assess what we do next.

tcarmet:
As discussed adding this feature is harder than expected and not worthy

PR #260, opened 2023-06-15
[PTFE-324] 🐛 Fix GCP VM instance tags

gaspardmoindrot:
I took the opportunity to add a few things, and I also came across another bug. When a Github runner was launched, it would throw an error because it was trying to follow the same path as the GCloud runners (due to the fact that everything is based on webhook reception). That has been fixed now!


srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
```diff
@@ -70,6 +71,10 @@ def update_vm_metadata(self, instance_name: str, metadata: dict):
 
             labels = instance.labels or {}
             for key, value in metadata.items():
+                if len(value) > 63:
+                    value = value[:63]
+                value = value.lower()
+                value = ''.join(c if c.islower() or c.isdigit() or c in ['_', '-'] else '_' for c in value)
```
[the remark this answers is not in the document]
tcarmet: Changing to dash instead of underscore

tcarmet (changes requested):

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
```diff
@@ -70,6 +71,9 @@ def update_vm_metadata(self, instance_name: str, metadata: dict):
 
             labels = instance.labels or {}
             for key, value in metadata.items():
+                value = value[:63]
+                value = value.lower()
+                value = ''.join(c if c.islower() or c.isdigit() or c in ['_', '-'] else '-' for c in value)
```
tcarmet, line 76: I have a better time understanding the code with a regex:
```suggestion
                value = re.sub(r'\W', '-', value)
```

Not to forget the proper import:
```python
import re
```

gaspardmoindrot: Wouldn't this be better? That way, we also keep the `_` when they are present. In reality, it's not that important:
`value = re.sub(r'[^a-z0-9_-]', '_', value)`

srcs/web/WebhookManager.py
```diff
@@ -92,15 +92,19 @@ def workflow_job(self, payload: WebHook):
...
 
-        runner_m.factory.cloud_manager.update_vm_metadata(
-            payload.workflow_job.runner_name,
-            dict(
-                status=status["status"],
-                repository=payload.repository.name,
-                workflow=payload.workflow_job.workflow_name,
-                job=payload.workflow_job.name
+        if (
+            payload.action != "queued"
```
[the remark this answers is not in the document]
gaspardmoindrot: I'll check!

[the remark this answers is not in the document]
gaspardmoindrot: Very insightful thinking on your part, indeed it was causing errors, and I have resolved it!

srcs/web/WebhookManager.py
```diff
@@ -92,15 +92,19 @@ def workflow_job(self, payload: WebHook):
...
+            and "gcloud" in payload.workflow_job.runner_name
```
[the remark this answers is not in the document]
gaspardmoindrot: I tried many things, and eventually, I realized that it simply required excluding GitHub runners from executing this function. I added a condition to prevent that. Let me know if you think it's good or if you prefer another approach. I find it quite consistent, and essentially, we want this function to always execute except for GitHub.

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
gaspardmoindrot: Done

srcs/web/WebhookManager.py
```diff
@@ -92,15 +92,20 @@ def workflow_job(self, payload: WebHook):
...
-                repository=payload.repository.name,
-                workflow=payload.workflow_job.workflow_name,
-                job=payload.workflow_job.name
-            )
-        )
+        if (
+            payload.action != "queued"
+            and payload.workflow_job.conclusion != "skipped"
+        ):
+            if payload.workflow_job.runner_group_name != "GitHub Actions":
```
tcarmet, line 99: It's not enough, you are going to get event for runners that are:
- Both not managed by the runner-manager and not GitHub hosted.
- Managed by another runner-manager.

tcarmet (changes requested):

srcs/web/WebhookManager.py
gaspardmoindrot: For "Managed by another runner-manager," it's not an issue as the function is also implemented for other runner-managers. Right? Even if nothing will happen
image

gaspardmoindrot: I have marked the PR as a draft because what I just added is not working. Just let me know if it is what you had in mind. I just can't figure out why, but sometimes the Redis is initialized and other times it's not. I will investigate while waiting!

gaspardmoindrot: Alternatively, I can use an `if` statement with `"self-hosted" in payload.workflow_job.labels`, but I have a feeling that it won't work every time since we don't always specify "self-hosted" as a tag for the runners.

srcs/runners_manager/runner/RedisManager.py
```diff
@@ -46,6 +46,16 @@ def get_runner(self, name: str) -> Runner:
...
+    def check_runners(self, manager_name: str) -> bool:
+        """
+        Check if a runner exist or not for a specific manager_name
+        """
+        runners: Runner = self.get_all_runners()
+        for runner in runners:
+            if manager_name == runner.name:
+                return True
+        return False
+
```
tcarmet, line 58: For a given manager name or for a given runner name?
```suggestion
    def check_runners(self, name: str) -> bool:
        """
        For a given runner name, check if it is managed by the service.
        """
        runners: Runner = self.get_all_runners()
        for runner in runners:
            if name == runner.name:
                return True
        return False

```

srcs/web/WebhookManager.py
```diff
@@ -92,11 +95,12 @@ def workflow_job(self, payload: WebHook):
             }
         )
 
+        logger.debug("Redis is initialize")
```
tcarmet, line 98:
```suggestion
        logger.debug("Redis is initialized")
```

srcs/web/app.py
```diff
@@ -24,6 +27,23 @@
...
 
+@lru_cache()
+def get_args() -> EnvSettings:
+    return EnvSettings()
+
+
+@lru_cache()
+def get_redis() -> RedisManager:
+    args: EnvSettings = get_args()
+    settings: dict = setup_settings(args.setting_file)
```
tcarmet, line 38: Why is setup_settings not setup as a `lru_cache`?

srcs/web/app.py
```diff
@@ -129,11 +149,15 @@ async def stop_runner(request: Request):
 
 
 @app.post("/webhook")
-async def webhook_post(data: WebHook, request: Request):
+async def webhook_post(
+    data: WebHook,
+    request: Request,
+    r: RedisManager = Depends(get_redis),
```
tcarmet, line 155:
```suggestion
    redis: RedisManager = Depends(get_redis),
```

srcs/web/app.py
```diff
@@ -129,11 +149,15 @@ async def stop_runner(request: Request):
...
+):
     """
     Webhook point for Github
     """
-    WebHookManager(payload=data, event=request.headers["X-Github-Event"])()
+    WebHookManager(r, payload=data, event=request.headers["X-Github-Event"])()
```
tcarmet, line 160:
```suggestion
    WebHookManager(redis, payload=data, event=request.headers["X-Github-Event"])()
```

srcs/runners_manager/runner/RedisManager.py
```diff
@@ -46,6 +46,16 @@ def get_runner(self, name: str) -> Runner:
         data = json.loads(self.redis.get(name))
         return Runner.fromJson(data)
 
+    def check_runners(self, manager_name: str) -> bool:
```
tcarmet, line 49: rename it to `is_managed` or `runner_exist` something a bit more precise

tcarmet (changes requested):

srcs/runners_manager/runner/RedisManager.py
gaspardmoindrot: You're right

srcs/web/app.py
gaspardmoindrot: I find that it made the code look messy. Nevertheless, I made the change. Let me know what you think of it. I've tested everything, and it works perfectly.

gaspardmoindrot:
I realized that with the previous merge I did on the main branch, it messed things up a bit in the PR. I fixed everything and tested it. Everything's good now!

nootal (changes requested):

srcs/runners_manager/runner/RedisManager.py
```diff
@@ -46,6 +46,16 @@ def get_runner(self, name: str) -> Runner:
...
+        """
+        Check if a runner exist or not for a specific manager_name
```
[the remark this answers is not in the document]
gaspardmoindrot: Outdated !

nootal (approved):

PR #262, opened 2023-06-22
PTFE-456: break loop when demanding new runner

Abubakarr99:
When a new runner is required the function need_new_runner is stucked in a while loop until a runner is created. We noticed if the max is reached we get a lot of error message. So i use and if condition instead of while. In any case the manage_runners function that calls the need_runner function is ran every 2 mins so this would atleast reduce the number of log entries

### DIFF
--- a/srcs/runners_manager/runner/RunnerManager.py
+++ b/srcs/runners_manager/runner/RunnerManager.py
@@ -77,8 +77,12 @@ class RunnerManager(object):
         It won't create a Runner if we are already at the maximum we want.
         In the case the max is set at 0 we don't have a maximum.
         """
+        offline_runners = len(self.filter_runners(lambda r: r.has_run))
+        online_runners = len(self.filter_runners(lambda r: r.is_online))
+        creating_runners = len(self.filter_runners(lambda r: r.is_creating))
         if 0 < self.vm_type.quantity["max"] <= len(self.runners):
             logger.info("Runner not created, already to much")
+            logger.info(f"{online_runners} runners online, {creating_runners} creating, {offline_runners} offline")
             return
 
         if not self.redis.get_manager_running():
--- a/srcs/runners_manager/runner/RunnerManager.py
+++ b/srcs/runners_manager/runner/RunnerManager.py
@@ -77,9 +77,9 @@ class RunnerManager(object):
         It won't create a Runner if we are already at the maximum we want.
         In the case the max is set at 0 we don't have a maximum.
         """
-        offline_runners = len(self.filter_runners(lambda r: r.has_run))
-        online_runners = len(self.filter_runners(lambda r: r.is_online))
-        creating_runners = len(self.filter_runners(lambda r: r.is_creating))
+        offline_runners = len(self.filter_runners(lambda runner: runner.has_run))
+        online_runners = len(self.filter_runners(lambda runner: runner.is_online))
+        creating_runners = len(self.filter_runners(lambda runner: runner.is_creating))
         if 0 < self.vm_type.quantity["max"] <= len(self.runners):
             logger.info("Runner not created, already to much")
             logger.info(f"{online_runners} runners online, {creating_runners} creating, {offline_runners} offline")


srcs/runners_manager/runner/Manager.py
```diff
@@ -135,7 +135,7 @@ def manage_runners(self):
                 manager.delete_runner(runner)
 
             # Create if it's still not enough
-            while self.need_new_runner(manager):
+            if self.need_new_runner(manager):
```
tcarmet, line 138: What's the return type/value of `need_new_runner` and why changing to `if` has an impact?

Also, the `while` must have been set for a reason, to keep creating runners while they are needed. Is there a behavior impact on the code?

Abubakarr99: the `need_new_runner` returns a bool. the impact is that the function would break out of the loop to avoid the  `create_runner`  which is called in the loop continously. the refresh endpoint would called the `runner_manager` method every two minutes. There is no impact in the sense that if the manager has reach `max` no runner would be created. It is the same in a `while` or `if` except with if we reduce the number of log entries. I was not able to reproduce the issue I commented in the ticket. Maybe I am missing something in orther to reporduce it.

tcarmet: I see, when trying to reproduce the issue, were you setup on aws or google cloud?

As it seems that the gcloud runner never faced this issue (as far as I know, or we just never noticed it).

Abubakarr99: I tried it on the gcloud runner. It is not an issue per say. Maybe due to the limited pool space in the AWS runner that is why the  `max` limit was reached allowing @gaspardmoindrot  to find the weird behavior. With the gcloud runner manager their is a many runners available at any moment so the `limit` is hardly reached. Maybe also we had the log entries but for a short period of time.

gaspardmoindrot: I find it strange that you couldn't reproduce the error (whether with GCloud or AWS). How did you test it? Did you set up a pool with max=3 and bombard it with workflows, for example?

Try checking if Redis is properly initialized. It would be really helpful to find the error, as there might be something more than just a need for runners without available slots.

I'm not necessarily a fan of replacing the `while` with an `if`. My thought is that if someone needs to launch, let's say, 15 jobs in parallel, they would have to wait for 15 multiplied by 2 minutes, which is 30 minutes. I believe we can find an alternative approach that doesn't impact developers as much. Let me know what you think, @Abubakarr99.

tcarmet, line 138: - [ ] Add log with the number of with their respective status.
```
15 runners online
2 runners offline
1 runner creating
```

tcarmet, line 138:
```suggestion
            while self.need_new_runner(manager):
```

gaspardmoindrot: Agree with Thomas

gaspardmoindrot (commented):
I agree with Thomas, something like this could be done for the logs:
```
15 runners online, 2 runners offline, 1 runner creating
```

srcs/runners_manager/runner/RunnerManager.py
```diff
@@ -77,8 +77,12 @@ def create_runner(self) -> None:
         It won't create a Runner if we are already at the maximum we want.
         In the case the max is set at 0 we don't have a maximum.
         """
+        offline_runners = len(self.filter_runners(lambda r: r.has_run))
+        online_runners = len(self.filter_runners(lambda r: r.is_online))
+        creating_runners = len(self.filter_runners(lambda r: r.is_creating))
```
gaspardmoindrot, line 82:
```suggestion
        offline_runners = len(self.filter_runners(lambda runner: runner.has_run))
        online_runners = len(self.filter_runners(lambda runner: runner.is_online))
        creating_runners = len(self.filter_runners(lambda runner: runner.is_creating))
```

gaspardmoindrot (approved):
LGTM 🛫
Just a minor comment for clarification.

tcarmet (approved):

PR #271, opened 2023-07-11
PTFE-573: install podman instead of docker

Abubakarr99:


templates/init_runner_script.sh
```diff
@@ -81,9 +81,9 @@ EOF'
 
 sudo subscription-manager repos --enable=rhel-7-server-extras-rpms --enable=rhel-7-server-optional-rpms
 sudo yum install -y docker
-elif [ "${LINUX_OS_VERSION}" = "8" ]
+elif [ "${LINUX_OS_VERSION}" = "8" || "${LINUX_OS_VERSION}" = "9" ]
 then
-sudo dnf install -y podman-docker
+sudo dnf install -y podman-docker podman
 DOCKER_SERVICE_START="no"
 elif [ "${LINUX_OS_VERSION}" = "9" ]
```
nootal, line 88: What about this part?

Abubakarr99: yeah I forgot.

nootal (approved):

templates/init_runner_script.sh
```diff
@@ -81,15 +81,10 @@ EOF'
 
 sudo subscription-manager repos --enable=rhel-7-server-extras-rpms --enable=rhel-7-server-optional-rpms
 sudo yum install -y docker
-elif [ "${LINUX_OS_VERSION}" = "8" ]
+elif [ "${LINUX_OS_VERSION}" = "8" || "${LINUX_OS_VERSION}" = "9" ]
 then
-sudo dnf install -y podman-docker
+sudo dnf install -y podman-docker podman
```
tcarmet, line 86: Optional: Might not be necessary as the package could be implicitly included, but doesn't hurt to explicitly mention it as well. 
```suggestion
sudo dnf install -y podman-docker
```

tcarmet (approved):

PR #279, opened 2023-07-18
PTFE-610: bugfix rollback spot instances

Abubakarr99:
Spot instances are still being spawn. This is because the check that obtains the value of the spot key does not work properly. It always returns true. This is a fix.

### DIFF
--- a/srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
+++ b/srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
@@ -120,7 +120,7 @@ class GcloudManager(CloudManager):
         automatic_restart = True
         provisioning_model = "STANDARD"
         instance_termination_action = "DEFAULT"
-        preemptible = bool(runner.vm_type.config.get("spot", False))
+        preemptible = str(runner.vm_type.config.get("spot", "")).lower() == "true"
         if preemptible:
             provisioning_model = "SPOT"
             automatic_restart = False
--- a/srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
+++ b/srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
@@ -1,6 +1,7 @@
 import logging
 import re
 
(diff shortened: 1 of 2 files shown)


gaspardmoindrot (approved):
LGTM, I test it, it works well

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
```diff
@@ -120,7 +120,7 @@ def configure_instance(
         automatic_restart = True
         provisioning_model = "STANDARD"
         instance_termination_action = "DEFAULT"
-        preemptible = bool(runner.vm_type.config.get("spot", False))
+        preemptible = str(runner.vm_type.config.get("spot", "")).lower() == "true"
```
tcarmet, line 123: I think this is due to how Python's `bool` method behaves, which will consider the following:
```python
>>> bool("false")
True
>>> bool("False")
True
```

Try using the [strtobool](https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool) method, see if it helps solving our issue.

```suggestion
# assuming the following import is made
from distutils.util import strtobool

        preemptible = strtobool(runner.vm_type.config.get("spot", False))
```

tcarmet (changes requested):

srcs/runners_manager/vm_creation/gcloud/GcloudManager.py
Abubakarr99: yeah this works

tcarmet: then let's apply it.

gaspardmoindrot:
I tested locally, and it works

tcarmet (approved):
LGTM, please ensure the code work as expected with `spot: false`

gaspardmoindrot:
> LGTM, please ensure the code work as expected with `spot: false`

Yeah, it works well!

PR #290, opened 2023-07-31
PTFE-670: kustomize runner-manager v2

Abubakarr99:


.github/workflows/test-deployment.yaml
```diff
@@ -0,0 +1,33 @@
+---
+name: test-deployment
+
```
tcarmet, line 3:
```suggestion

permissions:
  contents: read

```

manifests/base/redis/kustomization.yaml
```diff
@@ -0,0 +1,21 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+configMapGenerator:
+  - name: redis-config
+    envs:
+      - redis-config.env
```
tcarmet, line 8: not needed inside redis.
```suggestion
```

manifests/base/redis/redis-config.env
```diff
@@ -0,0 +1 @@
+REDIS_OM_URL=redis://redis:6379
```
tcarmet, line 1: should be moved inside `base/runner-manager`

manifests/base/redis/statefulset.yaml
```diff
@@ -0,0 +1,38 @@
...
+spec:
+  serviceName: redis
+  replicas: 1
+  template:
+    spec:
+      securityContext:
+        fsGroup: 1001
+      containers:
+        - name: redis
+          image: "redis/redis-stack:6.2.6-v9"
```
tcarmet, line 15: Same image, the difference is that there's no "developer tools" that are not needed for a server.
```suggestion
          image: "redis/redis-stack-server:6.2.6-v9"
```

manifests/base/runner-manager/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+  - name: runner-manager
+    newName: ghcr.io/scality/runner-manager
+    newTag: 0.6.3
```
tcarmet, line 7: check if an image is available, could be one set with a git sha. An old release will not make your environment work.
```suggestion
    newTag: <git sha>
```

manifests/base/runner-manager/worker/kustomization.yaml
```diff
@@ -0,0 +1,11 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+commonLabels:
+  app.kubernetes.io/name: runner-manager-worker
+  app.kubernetes.io/instance: runner-manager-worker
+  app.kubernetes.io/version: 0.6.3
```
tcarmet, line 8: This label should be added automatically, or removed.
```suggestion
```

manifests/base/runner-manager/main/deployment.yaml
```diff
@@ -0,0 +1,37 @@
...
+              valueFrom:
+                configMapKeyRef:
+                  key: REDIS_OM_URL
+                  name: redis-config
+          image: runner-manager
+          imagePullPolicy: IfNotPresent
+          name: runner-manager
+          ports:
+            - containerPort: 80
+              name: http
```
tcarmet, line 29: It's 8000 now
```suggestion
          ports:
            - containerPort: 8000
              name: http
```

manifests/base/runner-manager/main/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+commonLabels:
+  app.kubernetes.io/name: runner-manager
+  app.kubernetes.io/instance: runner-manager
+  app.kubernetes.io/version: 0.6.3
```
tcarmet, line 8: Should be added automatically or removed:
```suggestion
```

manifests/base/runner-manager/main/service.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: runner-manager
+  labels: {}
+spec:
+  type: ClusterIP
+  ports:
+    - port: 80
```
tcarmet, line 9: Let's keep to 8000 for consistency with the deployment.
```suggestion
    - port: 8000
```

manifests/base/runner-manager/worker/deployment.yaml
```diff
@@ -0,0 +1,35 @@
...
+              valueFrom:
+                configMapKeyRef:
+                  key: REDIS_OM_URL
+                  name: redis-config
+          image: runner-manager
+          imagePullPolicy: IfNotPresent
+          name: runner-manager
+          ports:
+            - containerPort: 80
+              name: http
```
tcarmet, line 27: No port in the worker:
```suggestion
```

.github/workflows/test-deployment.yaml
```diff
@@ -0,0 +1,33 @@
...
+  test-deployment:
+    needs: docker-build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: helm/kind-action@v1
+      - name: Edit the runner-manager image
+        working-directory: manifests/base/runner-manager
+        run: |
+          kustomize edit set image runner-manager=ghcr.io/scality/runner-manager:${{ github.sha }}
```
tcarmet, line 23: Can the set be done within the build command?

tcarmet (changes requested):

.github/workflows/test-deployment.yaml
Abubakarr99: I don't think so. I have searched how to override the image but I could not find a solution. I thought about parsing the image tag while building but that does not work too.

tcarmet:
```suggestion
          kustomize edit set image ghcr.io/scality/runner-manager=ghcr.io/scality/runner-manager:${{ github.sha }}
```

manifests/base/runner-manager/kustomization.yaml
tcarmet:
```suggestion
```

.github/workflows/test-deployment.yaml
Abubakarr99: it is not working as expected. this is because i run the edit before deploying which I do through kustomize build

.github/workflows/test-deployment.yaml
```diff
@@ -0,0 +1,44 @@
...
+  test-deployment:
+    needs: docker-build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: helm/kind-action@v1
+      - name: Edit the runner-manager image
+        working-directory: tests
+        run: |
+          cat <<EOF > kustomization.yaml
```
tcarmet, line 26: The previous method should have worked

Abubakarr99: I prefer this way. I think it looks cleaner. In fact I think it should be a file in tests folder

.github/workflows/test-deployment.yaml
```diff
@@ -0,0 +1,44 @@
+---
+name: test-deployment
+
+permissions:
+  contents: read
+
+on: pull_request
```
tcarmet, line 7: Nitpicking, feels weird to see on at a lower level
```suggestion

on: pull_request

permissions:
  contents: read

```

manifests/base/runner-manager/main/deployment.yaml
```diff
@@ -0,0 +1,55 @@
...
+              cpu: 250m
+              memory: 1Gi
+            requests:
+              cpu: 250m
+              memory: 1Gi
+          livenessProbe:
+            failureThreshold: 3
+            httpGet:
+              path: /_health
+              port: 8000
```
tcarmet, line 41:
```suggestion
              port: http
```

manifests/base/runner-manager/redis-config.env
```diff
@@ -0,0 +1 @@
+REDIS_OM_URL=redis://redis:6379
```
tcarmet, line 1: Just realized something here, doesn't have to be now, but we'll have to figure out, eventually, how to transform the redis service name here, I believe that something like this should be set:

```yaml
# kustomization.yaml

replacements:
- source:
    kind: Service
    name: redis
  targets:
    - select:
        kind: ConfigMap
        name: redis-config
      fieldPaths:
        - data.REDIS_OM_URL
      options:
        delimiter: /
        index: 2 # might be 1 if the empty one is ignored
    
```

This is also may mean we might have to move this file into the redis folder and set those inside the `manifests/base/redis/kustomization.yaml` file

manifests/base/runner-manager/main/deployment.yaml
```diff
@@ -0,0 +1,55 @@
...
+            initialDelaySeconds: 5
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 5
+          readinessProbe:
+            failureThreshold: 3
+            httpGet:
+              path: /_health
+              port: 8000
```
tcarmet, line 50:
```suggestion
              port: http
```

tcarmet (approved):

manifests/base/runner-manager/main/deployment.yaml
Abubakarr99: why is it http

tcarmet: Because you defined as http in the container specs [here](https://github.com/scality/runner-manager/pull/290/files/33de6879aca18d0ac038072ac579a0a60906b481#diff-92d2cf506a56a4f1702389dc6dc9c1dc94bb3fd3a87a1382775a5061dc4b8963R27)

Abubakarr99: ok thanks

PR #292, opened 2023-07-31
PTFE-699 bootstrap runner-groups API calls

tcarmet:


runner_manager/clients/github.py
```diff
@@ -0,0 +1,134 @@
...
+from githubkit import GitHub as GitHubKit
+from githubkit.response import Response
+from githubkit.rest import RestNamespace as RestNamespaceKit
+from githubkit.rest.actions import ActionsClient as ActionsClientKit
+from githubkit.rest.models import GitHubRestModel
+from githubkit.utils import UNSET, Missing, exclude_unset
+
+
+class RunnerGroup(GitHubRestModel):
+    id: Optional[int] = None
```
gaspardmoindrot, line 26: Are you sure you want to make it optional? It's not a problem at the moment, but it could potentially become one in the future if someone codes something using the `id`.

gaspardmoindrot (approved):
LGTM 🔥

runner_manager/clients/github.py
tcarmet: I think we can't set it as required, as when we'll create the group, we have to provide a runnergroup object to the API call but we won't have the id of the group at this time. If we set it as required we will not be able to create this resource.

PR #301, opened 2023-08-01
[PTFE-680] ✨ Support webhook endpoint

gaspardmoindrot:


runner_manager/main.py
```diff
@@ -21,3 +23,14 @@ def startup_event():
 @app.get("/_health")
 def health():
     return Response(status_code=200)
+
+
+@app.post("/webhook", status_code=202)
+async def webhook(request: Request):
```
tcarmet, line 29: We're not setting the project with async compatibility. Use regular methods.

tests/api/test_job.py
```diff
@@ -0,0 +1,12 @@
+def test_webhook_job_endpoint(client):
+    headers = {"X-GitHub-Event": "push"}
+    webhook = {
```
tcarmet, line 3: Define at least webhook as a pytest fixture.

runner_manager/main.py
```diff
@@ -21,3 +23,14 @@ def startup_event():
...
+    try:
+        event: WebhookEvent = parse(
+            request.headers["X-GitHub-Event"], await request.body()
+        )
+        return "Accepted"
```
tcarmet, line 34: return `Response` object.

runner_manager/main.py
tcarmet, line 29: We should be able to defined `WebhookEvent` directly, fastapi supports pydantic models to be defined as parameter in the function and will then automatically check if the request is valid as well as producing the proper openapi spec/docs.
```suggestion
async def webhook(webhook: WebhookEvent):
```

tcarmet (changes requested):
This task was initially defined as blocked by the api routes and we should encounter this method in the defined `webhook.py` file, you can chose to go ahead and then move this method or add the routes and merge this code later on.

PR #309, opened 2023-08-02
PTFE-693: unit test for settings class

Abubakarr99:


tests/unit/models/test_settings.py
```diff
@@ -0,0 +1,22 @@
...
+
+@fixture
+def test_env_settings():
+    env_settings = {
+        "NAME": "name-test",
+        "REDIS_OM_URL": "redis://localhost:6379/1",
+        "GITHUB_BASE_URL": "http://localhost:4999",
+    }
+    settings = Settings()
+    settings.Config.customise_sources(env_settings=env_settings, init_settings={}, file_secret_settings={})
```
tcarmet, line 19: customize source is implicitly called, no need to make direct call to it, if you want to test env settings you should set env variable like the following:
```python
import os

os.environ['NAME'] = 'name-test'

settings = Settings()

assert os.environ['NAME'] == 'name-test'

File: tests/unit/models/test_settings.py [hunk 2]
```diff
@@ -0,0 +1,22 @@
+from runner_manager.models.settings import Settings
+from pytest import fixture
+
+
+def test_settings_default_values():
+    settings = Settings()
+    assert settings.name == "runner-manager"
+    assert settings.redis_om_url == "redis://localhost:6379/0"
+    assert settings.github_base_url == "http://localhost:4010"
```
tcarmet, line 9: did I put that as default value? I'm surprised.

tests/unit/models/test_settings.py
```diff
@@ -0,0 +1,22 @@
...
+
+@fixture
```
tcarmet, line 11:
```suggestion
```

tcarmet (changes requested):

tests/unit/models/test_settings.py
Abubakarr99: I think without the @fixture it just picks up the value used by the settings passed in during the startup of application

tests/unit/models/test_settings.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def test_yaml_config(temp_yaml_file):
+    config_file = ConfigFile(config_file=Path(temp_yaml_file))
+    settings = Settings()
+    settings.__config__.config = config_file
+    config_data = yaml_config_settings_source(settings)
+    assert config_data["name"] == "test-runner-manager"
+    assert config_data["redis_om_url"] == "redis://localhost:6379/0"
+    assert config_data["github_base_url"] == "https://github.com"
```
tcarmet, line 41: This is the expected behavior from which we will be working with when using settings:
```suggestion
    os.environ['CONFIG_FILE'] = temp_yaml_file
    # settings should automatically look for the config file
    # environment variable and load the file if it exists
    settings = Settings()
    yaml_file = yaml.load(temp_yaml_file)
    assert settings["name"] ==  yaml_file["name"]
    assert settings["redis_om_url"] ==  yaml_file["redis_om_url"]
    assert settings["github_base_url"] == yaml_file["github_base_url"]
    
```

tests/unit/models/test_settings.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def test_env_file():
+    os.environ["REDIS_OM_URL"] = "redis://localhost:6379/0"
+    os.environ["GITHUB_BASE_URL"] = "https://github.com"
+    os.environ["NAME"] = "test-runner-manager"
+    settings = Settings()
+    assert settings.name == "test-runner-manager"
```
tcarmet, line 49:
```suggestion
    assert settings.name == os.getenv("NAME")
```

tests/unit/models/test_settings.py
```diff
@@ -0,0 +1,51 @@
...
+from pytest import fixture
+
+from runner_manager.models.settings import (
+    ConfigFile,
+    Settings,
+    yaml_config_settings_source,
+)
+
+
+@fixture
```
tcarmet, line 14: not a fixture:
```suggestion
```

tests/unit/models/test_settings.py
```diff
@@ -0,0 +1,51 @@
...
+def test_settings_default_values():
+    settings = Settings()
+    assert settings.name == "runner-manager"
+    assert settings.redis_om_url == os.getenv("REDIS_OM_URL")
```
tcarmet, line 18: If we don't have those var set it won't work.

And I don't think we should force user to set it, the way it was before with the hardcoded value was fine I believe.

tcarmet (changes requested):
Optionally add a testcase with errors expecting to be raised:

What happens when you do:
```python
Settings(redis_om_url="toto")
```
Can be tested like:
```

with pytest.raises(<expected exception>):
    Settings(redis_om_url="toto")

```

tests/unit/models/test_settings.py
```diff
@@ -0,0 +1,55 @@
...
+import yaml
+from pytest import fixture
+
+from runner_manager.models.settings import Settings
+
+
+def test_settings_default_values():
+    settings = Settings()
+    assert settings.name == "runner-manager"
+    assert settings.redis_om_url == os.getenv("REDIS_OM_URL")
```
tcarmet, line 13:
```suggestion
    assert settings.redis_om_url == os.getenv("REDIS_OM_URL", "redis://localhost:6379/0")
```

tests/unit/models/test_settings.py
```diff
@@ -0,0 +1,55 @@
...
+def test_invalid_redis_url():
+    with pytest.raises(ValueError):
+        Settings(redis_om_url="invalid_redis_url")
+
+def test_invalid_github_url():
+    with pytest.raises(ValueError):
+        Settings(github_base_url="invalid_github_url")
+
+def test_yaml_config(temp_yaml_file):
+    os.environ["CONFIG_FILE"] = temp_yaml_file
```
tcarmet, line 37:
```suggestion
    os.environ["CONFIG"] = temp_yaml_file
```

tests/unit/models/test_settings.py
```diff
@@ -0,0 +1,55 @@
...
+def test_invalid_redis_url():
+    with pytest.raises(ValueError):
+        Settings(redis_om_url="invalid_redis_url")
+
+def test_invalid_github_url():
+    with pytest.raises(ValueError):
+        Settings(github_base_url="invalid_github_url")
+
+def test_yaml_config(temp_yaml_file):
+    os.environ["CONFIG"] = temp_yaml_file
```
tcarmet, line 37:
```suggestion
    os.environ["CONFIG_FILE"] = temp_yaml_file
```

tests/unit/models/test_settings.py
tcarmet: tested and actually the value is set to None if the value is not set so the test is good.

tests/unit/models/test_settings.py
```diff
@@ -0,0 +1,55 @@
...
+    redis_om_url: redis://localhost:6379/0
+    github_base_url: https://github.com
+    """
+    yaml_file = tempfile.NamedTemporaryFile(mode="w", delete=False)
+    yaml_file.write(yaml_data)
+    return yaml_file.name
+
+def test_invalid_redis_url():
+    with pytest.raises(ValueError):
+        Settings(redis_om_url="invalid_redis_url")
```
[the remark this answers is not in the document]
tcarmet: Already done, had to add the same config in the trunk.yaml.

tcarmet (approved):

PR #311, opened 2023-08-02
Fix deployment tests BadRequest failure 

tcarmet:
Deployment tests failing with the following error:

```shell
configmap/redis-config-6tc26t8d82 created
service/redis created
service/runner-manager created
deployment.apps/runner-manager-worker created
statefulset.apps/redis created
Error from server (BadRequest): error when creating "STDIN": Deployment in version "v1" cannot be handled as a Deployment: json: cannot unmarshal number into Go struct field Container.spec.template.spec.containers.args of type string
```

Error due to setting `- 0.0.0.0` instead of `- '0.0.0.0'` as args on the container. Fixed it by removing the args/command and using the default values to run the runner-manager API.

Couple of extra changes made along the way while debugging this issue:

- Changed the kustomize image override.
- Added kubernetes vscode extension.
- Small lint/format fix.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -27,7 +27,8 @@
         "GitHub.copilot",
         "bierner.markdown-mermaid",
         "tamasfe.even-better-toml",
-        "github.vscode-github-actions"
+        "github.vscode-github-actions",
+        "ms-kubernetes-tools.vscode-kubernetes-tools"
       ],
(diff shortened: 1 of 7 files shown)


gaspardmoindrot (approved):

Abubakarr99:
Yeah it seems like checkcov is not consistent on that one.

Abubakarr99 (approved):

PR #314, opened 2023-08-03
PTFE-673: logger for runner manager

Abubakarr99:


runner_manager/logging.py
```diff
@@ -0,0 +1,20 @@
+import logging
+
+from runner_manager.models.settings import Settings
+
+settings = Settings()
+
+log_level = getattr(logging, settings.log_level.upper(), logging.INFO)
```
tcarmet, line 7: Check exactly what is the type of logging.INFO and all the others,
You may be able to do a specific type definitions in settings allowing you to remove the need of applying the transformation here.

Or apply this transformation directly in settings with the Field function

runner_manager/logging.py
```diff
@@ -0,0 +1,20 @@
+import logging
+
+from runner_manager.models.settings import Settings
+
+settings = Settings()
```
tcarmet, line 5: use get_settings to use the cache

runner_manager/logging.py
```diff
@@ -0,0 +1,20 @@
...
+
+log_level = getattr(logging, settings.log_level.upper(), logging.INFO)
+log = logging.getLogger(settings.name)
+log.setLevel(log_level)
+
+# console handler
+console_handler = logging.StreamHandler()
```
tcarmet, line 12: What does it do?

tcarmet (changes requested):

runner_manager/logging.py
Abubakarr99: logging.INFO is a int.

runner_manager/logging.py
Abubakarr99: it adds the handler to log the console stdout

runner_manager/logging.py
Abubakarr99: I'm not sure if this is what you want but i modified it
`log_level: str = Field(default="INFO", pattern=r'^[A-Z]+$')`

tcarmet (approved):

runner_manager/logging.py
tcarmet: Otherwise you are reading the configs all over again each time you call it
```suggestion
from runner_manager.dependencies import get_settings

settings = get_settings()
```

tcarmet:
Think about adding a new test case to ensure that your code is covered, a simple call should do:

```python
# tests/unit/test_logging.py

from runner_manager.logging import log

def test_log():
    log.info('info')
```

PR #315, opened 2023-08-03
PTFE-667: code coverage

Abubakarr99:


.github/workflows/tests.yaml
```diff
@@ -29,7 +29,9 @@ jobs:
           cache: poetry
       - run: poetry install
       - name: Run tests
-        run: poetry run pytest tests/${{ matrix.test }}
+        run: poetry run pytest tests/${{ matrix.test }} --cov=tests/${{ matrix.test }} --cov-report=xml
+      - name: Upload coverage
+        uses: codecov/codecov-action@v2
```
tcarmet, line 34: Not sure it will make a huge difference but seems harmless:
```suggestion
        uses: codecov/codecov-action@v2
        if: always()
```

pyproject.toml
```diff
@@ -25,6 +25,7 @@ mkdocs-material = "^9.1.21"
 [tool.poetry.group.dev.dependencies]
 connexion = "^2.14.2"
 pytest = "^7.4.0"
+pytest-cov = "^4.0.0"
 hypothesis = "^6.82.0"
```
tcarmet, line 30: Can you try adding this here:
```toml
[tool.pytest]
adopt = --cov-report=xml # add remaining --cov flags
```

The goal is to remove the need to have those flags in the command line, making it implicit to generate code coverage reports.


Not 100% sure on the exact syntax so feel free to check pytest-cov documentation.

.github/workflows/tests.yaml
```diff
@@ -29,7 +29,9 @@ jobs:
           cache: poetry
       - run: poetry install
       - name: Run tests
-        run: poetry run pytest tests/${{ matrix.test }}
+        run: poetry run pytest tests/${{ matrix.test }} --cov=tests/${{ matrix.test }} --cov-report=xml
```
tcarmet, line 32: You sure about the cov flag? Seems weird
```suggestion
        run: poetry run pytest tests/${{ matrix.test }} --cov=runner_manager --cov-report=xml
```

tcarmet (changes requested):
Check if we're not bound to commit coverage files,
Per example adding `.coverage` (where the coverage files are stored) in the `.gitignore` and `.dockerignore` file

.github/workflows/tests.yaml
Abubakarr99: yeah since the test suite are seperate each has its own workflow so to speak. with your suggestion the code coverage will be 0% for api if i am running the test on `tests/unit` for example

pyproject.toml
tcarmet: modified the suggested command with stuff that I found on the internet.

.github/workflows/tests.yaml
tcarmet: The goal is not to see the coverage of the tests, but the coverage of the code itself. We want to know what is covered within our code.

.github/workflows/tests.yaml
```diff
@@ -29,7 +29,9 @@ jobs:
           cache: poetry
       - run: poetry install
       - name: Run tests
-        run: poetry run pytest tests/${{ matrix.test }}
+        run: poetry run ${{ matrix.test }}
```
tcarmet, line 32:
```suggestion
        run: poetry run tests/${{ matrix.test }}
```

tcarmet: Actually, to be ignored given your addition of scripts.

pyproject.toml
```diff
@@ -40,3 +41,7 @@ stubPath = "./typings"
 reportMissingImports = false
 reportMissingModuleSource = false
 typeCheckingMode = "basic"
+
+[tool.poetry.scripts]
+unit = "pytest --cov=tests/unit --cov-report=xml tests/unit"
+api = "pytest --cov=tests/api --cov-report=xml tests/api"
```
tcarmet, line 47:
```suggestion

[tool.pytest.ini_options]
addopts = "--cov=runner_manager --cov-report=xml"

[tool.poetry.scripts]
unit = "pytest tests/unit"
api = "pytest tests/api"
```

pyproject.toml
```diff
@@ -34,4 +35,22 @@ yamllint = "^1.32.0"
...
+
+[tool.pyright]
+include = ["runner_manager"]
+exclude = ["**/__pycache__", "__pypackages__", "**/node_modules", "./dist"]
+venv = ".venv"
+venvPath = "."
+stubPath = "./typings"
+reportMissingImports = false
+reportMissingModuleSource = false
+typeCheckingMode = "basic"
```
tcarmet, line 48: Should be removed, those settings are now located on `pyrightconfig.json`
```suggestion
```

pyproject.toml
```diff
@@ -34,4 +35,22 @@ yamllint = "^1.32.0"
 
 [build-system]
 requires = ["poetry-core"]
-build-backend = "poetry.core.masonry.api"
```
tcarmet, line 37: What's the reason to remove this one? Should be set back.

tcarmet (changes requested):

Abubakarr99:
The suggestion you provided @tcarmet  did not work it is meant for running scripts that are in the project which is not the case for pytest. Another way is to also add option testpaths in the poetry.toml file 
`testpaths = ["tests/api", "tests/unit"]` and with poetry run pytest in runs the test on the test paths. This would mean we would have to modify the workflow to only run once and not in a matrix and I'm not sure that is what we want maybe it is.

tcarmet (approved):

tcarmet:
> The suggestion you provided @tcarmet did not work it is meant for running scripts that are in the project which is not the case for pytest. Another way is to also add option testpaths in the poetry.toml file `testpaths = ["tests/api", "tests/unit"]` and with poetry run pytest in runs the test on the test paths. This would mean we would have to modify the workflow to only run once and not in a matrix and I'm not sure that is what we want maybe it is.

Up to you, you can leave it as it was with: `poetry run pytest tests/${{ matrix.test }}` but your change with `${{ matrix.test.name }}` and `${{ matrix.test.path }}` work as well and might come useful in the future so LGTM.

PR #329, opened 2023-08-08
PTFE-685: redis tests

Abubakarr99:


tests/unit/conftest.py
```diff
@@ -15,16 +16,12 @@ def settings() -> Settings:
...
-    """Flush redis before tests."""
-
-    redis_connection = get_redis_connection(
-        url=settings.redis_om_url, decode_responses=True
-    )
-    redis_connection.flushall()
-
+def redis():
+    redis_my_proc = factories.redis_proc()
+    redis_connection = factories.redisdb(redis_my_proc)
```
tcarmet, line 21: If we can't setup as a url:
```suggestion
    redis_connection = factories.redisdb(redis_my_proc)

    # Clean up if needed
    class Meta
        database = redis_connection
    
    BaseModel.meta.database = redis_connection
```

tcarmet (commented):
For the unit tests, given that you are setting up redis as a proc and we should no longer need the redis available the following should work:
```shell

docker compose stop # Stop all containers from running
poetry run pytest tests/unit
```

tcarmet (approved):

PR #332, opened 2023-08-08
PTFE-679 implementing webhook endpoint and authentication

tcarmet:


runner_manager/routers/webhook.py
```diff
@@ -1,8 +1,42 @@
-from fastapi import APIRouter, Response
+from typing import Annotated
+
+from fastapi import APIRouter, Depends, Header, HTTPException
```
tcarmet, line 3:
```suggestion
from fastapi import APIRouter, Depends, Header, HTTPException
from fastapi.security import Security
```

PR #334, opened 2023-08-09
PTFE-675 GCP backend

Abubakarr99:


runner_manager/backend/gcloud.py
```diff
@@ -0,0 +1,175 @@
...
+    def image_client(self) -> ImagesClient:
+        """Returns a GCP Image client."""
+        return ImagesClient()
+
+    @property
+    def zone_operation_client(self) -> ZoneOperationsClient:
+        """Returns a GCP Zone Operation client."""
+        return ZoneOperationsClient()
+
+    def configure_instance(self, runner: Runner):
```
tcarmet, line 41: This function should not exist anymore most of those operation and variable definition can be set as default value of the GCPInstanceConfig, you should be able to pass GCPInstanceConfig to `instance_resource` parameter.

Check if you can inherit from it `class GCPInstanceConfig(InstanceConfig, Instance):` and override some of the var, or define as default the one you need.

tests/unit/backend/test_gcp.py
```diff
@@ -0,0 +1,58 @@
...
+@fixture()
+def gcp_group() -> RunnerGroup:
+    runner_group: RunnerGroup = RunnerGroup(
+        id=2,
+        name="test",
+        organization="test",
+        backend=GCPBackend(
+            name=Backends.gcloud,
+            config=GCPConfig(
+                project_id=os.getenv("GCP_PROJECT_ID"),
```
tcarmet, line 21: You should follow the same env var as the gcloud cli for consistency and not invent your own:
`CLOUDSDK_CORE_PROJECT`
`CLOUDSDK_COMPUTE_ZONE`

This way setting those will also set the one interpreted by `gcloud`

Additionally setting those var as `Field` in the model could allow you to retrive those from the env directly:

```python

from pydantic import Field

class GCPConfig(BackendConfig):
    """Configuration for GCP backend."""

    project_id: str = Field(..., env="CLOUDSDK_CORE_PROJECT")
    zone: str = Field(..., env="CLOUDSDK_COMPUTE_ZONE")
    service_account_email: str = "default"


config = GCPConfig()
config.zone # should return the zone from the env var if set.
```

tcarmet (changes requested):

tests/unit/backend/test_gcp.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@fixture()
+def gcp_runner(runner: Runner, gcp_group: RunnerGroup) -> Runner:
+    # Cleanup and return a runner for testing
+    gcp_group.backend.delete(runner)
+    return runner
+
+
+@mark.skipif(
```
tcarmet, line 48: Checkout if you can use markers to apply the same logic:
https://docs.pytest.org/en/6.2.x/example/markers.html#marking-platform-specific-tests-with-pytest

runner_manager/backend/gcloud.py
```diff
@@ -0,0 +1,175 @@
...
+from runner_manager.models.runner import Runner
+
+
+class GCPBackend(BaseBackend):
+    name: Literal[Backends.gcloud] = Field(default=Backends.gcloud)
+    config: GCPConfig
+    instance_config: GCPInstanceConfig
+
+    @property
+    def compute_client(self) -> InstancesClient:
```
tcarmet, line 26: Would rather see this one listed as client for consistency with other backends
```suggestion
    def client(self) -> InstancesClient:
```

tests/unit/backend/test_gcp.py
tcarmet: my bad, then in this case let's change BackendConfig to inherit from BaseSettings in this case it will take env variable then you can do:
```python
class GCPConfig(BackendConfig):
    """Configuration for GCP backend."""

    project_id: str = Field(..., alias="CLOUDSDK_CORE_PROJECT")
    zone: str = Field(..., alias="CLOUDSDK_COMPUTE_ZONE")
    service_account_email: str = "default"
```

tests/unit/backend/test_gcp.py
tcarmet: It's ok, let's leave it as is for now, we'll review once the backend config story is figured out.

tcarmet (changes requested):

runner_manager/models/backend.py
```diff
@@ -46,3 +53,56 @@ class DockerConfig(BackendConfig):
...
+
+class GCPConfig(BackendConfig):
+    """Configuration for GCP backend."""
+
+    project_id: str
+    zone: str
+    service_account_email: str = "default"
+
+
+class GCPInstanceConfig(InstanceConfig):
```
tcarmet, line 66: No success in doing this?
```suggestion
class GCPInstanceConfig(InstanceConfig, Instance):
```

runner_manager/models/backend.py
```diff
@@ -46,3 +53,56 @@ class DockerConfig(BackendConfig):
...
     base_url: str = "unix:///var/run/docker.sock"
+
+
+class GCPConfig(BackendConfig):
+    """Configuration for GCP backend."""
+
+    project_id: str
+    zone: str
+    service_account_email: str = "default"
+
```
tcarmet, line 64: For ref how the backend config will look like in the config file.
```yaml
- name: toto
  backend:
    name: gcloud
    config:
      project_id: tata
      zone: titi
    instance_config:
      <instance config> 
```

runner_manager/models/backend.py
```diff
@@ -46,3 +53,56 @@ class DockerConfig(BackendConfig):
...
+                "network": f"projects/{gcp_config.project_id}/global/networks/{network}",
+                "access_configs": [
+                    {
+                        "name": "External NAT",
+                        "type_": "ONE_TO_ONE_NAT",
+                    }
+                ],
+            }
+        ]
+        self.instance.labels = labels or {}
```
tcarmet, line 108:
```suggestion
    image_family: str
    image_project: str
    machine_type: str
    network: str
    labels: Optional[Dict[str, str]] = None
    
    image: Optional[str] = None
    
    disks: Optional[List[AttachedDisk]] = None
    netwok_interfaces: Optional[List[NetworkInterfaces]] = None
    
    
    def configure_instance(self, runner: Runner) -> Instance:
        return Instance(
            name=runner.name,
            # add remaining params from self)
```

runner_manager/backend/gcloud.py
```diff
@@ -0,0 +1,126 @@
...
+            self.wait_for_operation(
+                project=self.config.project_id,
+                zone=self.config.zone,
+                operation=ext_operation.name,
+            )
+            runner.instance_id = runner.name
+
+        except Exception as e:
+            raise e
+        return super().create(runner)
```
tcarmet, line 74:
```suggestion
    def create(self, runner: Runner):
        try:
            image = self.get_image()
            disks = self.get_disks()
            self.instance_config.image = image
            self.instance_config.disks = disks
            instance: Instance = self.config_instance.configure_instance(runner)
            ext_operation: ExtendedOperation = self.compute_client.insert(
                project=self.config.project_id,
                zone=self.config.zone,
                instance_resource=instance,
            )
            self.wait_for_operation(
                project=self.config.project_id,
                zone=self.config.zone,
                operation=ext_operation.name,
            )
            runner.instance_id = runner.name

        except Exception as e:
            raise e
        return super().create(runner)
```

runner_manager/models/backend.py
Abubakarr99: yeah.

runner_manager/backend/gcloud.py
Abubakarr99: there are other clients too. Or maybe be we don't set them at all and call them when needed

like `compute_v1.InstanceClient()`

tcarmet: the compute_v1.InstanceClient is the same as InstanceClient or at least it should be, it's just versionning made inside the library in case there are introducing breaking changes. 

Setting just this one as client is enough, and the other image_client and stuff like this you can leave as is.

runner_manager/models/backend.py
```diff
@@ -46,3 +47,36 @@ class DockerConfig(BackendConfig):
...
+
+    def configure_instance(self, runner: Runner) -> Instance:
+        """Configure instance."""
+        return Instance(
+            name=runner.name,
+            disks=self.disks,
+            machine_type=self.machine_type,
+            network_interfaces=self.network_interfaces,
+            labels=self.labels,
+        )
```
tcarmet, line 82: Pretty cool, it looks cleaner now IMHO.

runner_manager/models/backend.py
```diff
@@ -46,3 +47,36 @@ class DockerConfig(BackendConfig):
...
+
+    project_id: str
+    zone: str
+    service_account_email: str = "default"
+
+
+class GCPInstanceConfig(InstanceConfig):
+    image_family: str
+    image_project: str
+    machine_type: str
```
tcarmet, line 63: I wouldn't mind having some default values:
```suggestion
    image_family: str = "ubuntu-2004-lts"
    image_project: str = "ubuntu-os-cloud"
    machine_type: str = "e2-small"
```

tcarmet (approved):

runner_manager/backend/gcloud.py
Abubakarr99: ok

runner_manager/models/backend.py
```diff
@@ -46,3 +47,36 @@ class DockerConfig(BackendConfig):
...
     base_url: str = "unix:///var/run/docker.sock"
+
+
+class GCPConfig(BackendConfig):
+    """Configuration for GCP backend."""
+
+    project_id: str
+    zone: str
+    service_account_email: str = "default"
+
```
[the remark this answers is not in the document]
Abubakarr99: The pydantic Fields is still not working. with the alias it is asking for `CLOUDSDK_CORE_PROJECT` instead of `project_id`

runner_manager/models/backend.py
Abubakarr99: ok

runner_manager/models/backend.py
Abubakarr99: yeah

runner_manager/models/backend.py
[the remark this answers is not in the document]
Abubakarr99: i also can not get pytest to skip by passing the `not.gcp_group.backend.google_application_credentials` is seems to not be available

PR #337, opened 2023-08-10
PTFE-698 Add fastapi dependency for GitHub client 

tcarmet:


runner_manager/models/settings.py
```diff
@@ -34,16 +35,62 @@ class LogLevel(str, Enum):
...
     name: str = "runner-manager"
     redis_om_url: Optional[RedisDsn] = None
-    github_base_url: Optional[AnyHttpUrl] = None
     api_key: Optional[SecretStr] = None
     allowed_hosts: Optional[Sequence[str]] = [
         "localhost",
+        "127.0.0.1",
         "testserver",
+        "10.*",
     ]
```
tcarmet, line 44:
```suggestion
    allowed_hosts: Optional[Sequence[str]] = [
        "*",
    ]
```

PR #341, opened 2023-08-14
PTFE-719: runner group startup

Abubakarr99:


runner_manager/jobs/runner_groups.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def create_runner_group(settings: Settings = get_settings()):
+    runner_groups_configs = settings.runner_groups
+    existing_groups = RunnerGroup.find().all()
+    for runner_group_config in runner_groups_configs:
+        if runner_group_config.name not in [group.name for group in existing_groups]:
+            runner_group = runner_group_config.save()
+        else:
+            runner_group = runner_group_config.update()
```
tcarmet, line 13: The one that you should update is the existing one, not the one defined in settings.

```suggestion
            runner_group = RunnerGroup.find(RunnerGroup.name == runner_group_config.name).first()
            runner_group.update(runner_group_config)
```

runner_manager/models/settings.py
```diff
@@ -42,7 +42,7 @@ class Settings(BaseSettings):
     ]
     github_webhook_secret: Optional[SecretStr] = None
     log_level: LogLevel = LogLevel.INFO
-    runner_groups: List[BaseRunnerGroup] = []
+    runner_groups: List[RunnerGroup] = []
```
tcarmet, line 45: Why?

tcarmet (changes requested):
it's looking pretty good, but it's lacking a couple of things:

- enqueue the job on startup (check the main file)
- Add unit tests for the job

runner_manager/models/settings.py
Abubakarr99: I made a mistake I thihn `runner_group` is a list of RunnerGroup and not BaseRunnerGroup. Maybe I am wrong.

tcarmet: BaseRunnerGroup might not be needed anymore, I believe it was originally created to deal with the circular dependency issue. But while it exists, I think it's best we set runner_groups as `BaseRunnerGroup`

Abubakarr99: ok

runner_manager/jobs/runner_groups.py
```diff
@@ -0,0 +1,18 @@
...
+    for runner_group_config in runner_groups_configs:
+        if runner_group_config.name not in [group.name for group in existing_groups]:
+            runner_group = RunnerGroup(**runner_group_config.dict())
+            runner_group.save()
+        else:
+            runner_group = RunnerGroup.find(RunnerGroup.name == runner_group_config.name).first()
+            runner_group.update()
+            existing_groups.remove(runner_group)
+    for runner_group in existing_groups:
+        runner_group.delete()
```
tcarmet, line 18: Couple of things missing, we can work on it as needed.

- [ ] Add a `@classmethod` into the RunnerGroup called `create_group` or whatever name fits that will:

  - Take `cls` and `GitHub` client as parameter
  ```python
  @classmethod
  def create_group(cls, github: GitHub) -> "RunnerGroup":
      
  ```
  - Create a runner_group on github (or/and check if the runner group does not exist already).
  - Set the `RunnerGroup.id` to the retrieved id from the group creation.
  - `.save()` the runner
  - return the created runner group: `return cls.save()` works.
- [ ] Tests
- [ ] Modify inputs of method, with `settings: Settings, github: GitHub`
- [ ] When deleting the runner group we should delete existing runners (if any) and delete the group from github. Another class method can be created for it.

tcarmet (changes requested):

tcarmet:
Ended up taking the work of this PR, and applying my own review.

runner_manager/models/settings.py
```diff
@@ -39,6 +39,7 @@ class Settings(BaseSettings):
     allowed_hosts: Optional[Sequence[str]] = [
         "*",
     ]
+    healthcheck_interval: int = 60
```
tcarmet, line 42: Not so sure that I'll keep this configuration settings, I migth wait for the work on the group healthcheck.

tcarmet (approved):

PR #349, opened 2023-08-16
PTFE-674 aws backend

Abubakarr99:


Abubakarr99:
@tcarmet I am getting circular import errors in this PR. Please take a look.

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,81 @@
...
+        try:
+            instance = self.client.describe_instances(InstanceIds=[instance_id])
+            instance_id = instance["Instances"][0]["InstanceId"]
+        except ClientError as e:
+            if e.response["Error"]["Code"] == "InvalidInstanceID.NotFound":
+                log.error(f"Instance {instance_id} not found.")
+            else:
+                raise e
+        return Runner.find(Runner.instance_id == instance_id).first()
+
```
tcarmet, line 63: Is it on purpose that there's no update method (It's fine btw we don't have to implement it right away.)

runner_manager/models/backend.py
```diff
@@ -81,3 +81,42 @@ def configure_instance(self, runner: Runner) -> Instance:
...
+class AWSConfig(BackendConfig):
+    """Configuration for AWS backend."""
+
+    region: str = "us-west-2"
+
+
+class AWSInstanceConfig(InstanceConfig):
+    """Configuration for AWS backend instance."""
+
+    image: str = "ami-785db401"
```
tcarmet, line 95: Little comment telling us which image is this can be helpful.

tcarmet (changes requested):
Add `poetry add 'boto3-stubs[ec2]'` to help with pyright raising the missing stubfiles issue.

Then check if you can stop with this kind of syntax:
```python
for instance in instances["Reservations"]
```
And do something more like this:
```python
for instance in instances.reservations
```

Note that I'm still learning in this python typing story so we can spare some time together to dig into this issue.

Documentation: https://youtype.github.io/boto3_stubs_docs/mypy_boto3_ec2/

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,81 @@
...
+
+class AWSBackend(BaseBackend):
+    name: Literal[Backends.aws] = Field(default=Backends.aws)
+    config: AWSConfig
+    instance_config: AWSInstanceConfig
+
+    @property
+    def client(self) -> client:
+        """Returns a AWS Compute Engine client."""
```
nootal, line 20:
```suggestion
        """Return a AWS Compute Engine client."""
```
Make sure it's consistent elsewhere (either put 's' or don't)

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def create(self, runner: Runner) -> Runner:
+        """Create a runner."""
+        labels: Dict[str, str] = {}
+        if self.manager:
+            labels["runner-manager"] = self.manager
+        instance_resource: Dict = self.instance_config.configure_instance(runner)
+        try:
+            instance = self.client.run_instances(**instance_resource)
+            runner.instance_id = instance["Instances"][0]["InstanceId"]
```
nootal, line 31: Don't put several things in the try statement if you want to catch only things from the first one

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def delete(self, runner: Runner):
+        """Delete a runner."""
+        try:
+            if runner.instance_id:
+                self.client.terminate_instances(
+                    InstanceIds=[runner.instance_id]
+                )
+            else:
+                self.client.terminate_instances(InstanceIds=[runner.name])
```
nootal, line 44:
```suggestion
        runner_id = runner.instance_id if runner.instance_id else runner.name
        try:
            self.client.terminate_instances(InstanceIds=[runner_id])
```

You should put the if statement before the try statement to get the id.

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,81 @@
...
+                log.error(f"Instance {runner.instance_id} not found.")
+            else:
+                raise e
+        return super().delete(runner)
+
+    def get(self, instance_id: str) -> Runner:
+        """Get a runner."""
+        try:
+            instance = self.client.describe_instances(InstanceIds=[instance_id])
+            instance_id = instance["Instances"][0]["InstanceId"]
```
nootal, line 56: Same, don't put everything in try

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,81 @@
...
+        runners: List[Runner] = []
+        try:
+            instances = self.client.describe_instances()
+            for instance in instances["Reservations"]:
+                for i in instance["Instances"]:
+                    labels = i["Tags"]
+                    if self.manager and "runner-manager" in labels:
+                        runner = Runner(
+                            name=i["InstanceId"],
+                            instance_id=i["InstanceId"],
```
nootal, line 75:
```suggestion
            instances = self.client.describe_instances()
            for reservations in instances["Reservations"]:
                for instance in reservations["Instances"]:
                    labels = instance["Tags"]
                    if self.manager and "runner-manager" in labels:
                        runner = Runner(
                            name=instance["InstanceId"],
                            instance_id=instance["InstanceId"],
```
Be precise with the variable naming.

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def list(self) -> List[Runner]:
+        """List runners."""
+        runners: List[Runner] = []
+        try:
+            instances = self.client.describe_instances()
+            for instance in instances["Reservations"]:
+                for i in instance["Instances"]:
+                    labels = i["Tags"]
+                    if self.manager and "runner-manager" in labels:
```
nootal, line 72: self.manager does not seem to depend on the current instance or labels, and it's still a condition. So I would put it before the whole try thing, and then just return an empty list if it's not there.

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,81 @@
...
+            if e.response["Error"]["Code"] == "InvalidInstanceID.NotFound":
+                log.error(f"Instance {instance_id} not found.")
+            else:
+                raise e
+        return Runner.find(Runner.instance_id == instance_id).first()
+
+    def list(self) -> List[Runner]:
+        """List runners."""
+        runners: List[Runner] = []
+        try:
```
nootal, line 67: Too big try statement. You want to catch errors on the line where they could happen, not on some list loop

runner_manager/models/backend.py
```diff
@@ -81,3 +81,42 @@ def configure_instance(self, runner: Runner) -> Instance:
             network_interfaces=self.network_interfaces,
             labels=self.labels,
         )
+
+
+class AWSConfig(BackendConfig):
+    """Configuration for AWS backend."""
+
+    region: str = "us-west-2"
```
nootal, line 89: We want this to be configurable.

runner_manager/models/backend.py
nootal, line 95: We want this to be configurable.

nootal (changes requested):

runner_manager/backend/aws.py
Abubakarr99: just updated it

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def update(self, runner: Runner) -> Runner:
+        """Update a runner."""
+        try:
+            if runner.instance_id:
+                self.client.describe_instances(
+                    InstanceIds=[runner.instance_id],
+                    Filters=[
+                        {"Name": "instance-state-name", "Values": ["running"]},
+                        {"Name": "tag:runner-manager", "Values": ["aws"]},
```
tcarmet, line 131:
```suggestion
                        {"Name": "runner-manager", "Values": [self.manager]},
```

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,151 @@
...
+        return runners or []
+
+    def update(self, runner: Runner) -> Runner:
+        """Update a runner."""
+        try:
+            if runner.instance_id:
+                self.client.describe_instances(
+                    InstanceIds=[runner.instance_id],
+                    Filters=[
+                        {"Name": "instance-state-name", "Values": ["running"]},
```
tcarmet, line 130: Is it needed?
```suggestion
```

runner_manager/models/backend.py
```diff
@@ -82,3 +82,50 @@ def configure_instance(self, runner: Runner) -> Instance:
...
+    min_count: int = 1
+    user_data: Optional[str] = ""
+    block_device_mappings: Optional[List[Dict[str, str]]] = []
+    labels: Optional[Dict[str, str]] = None
+
+    class Config:
+        arbitrary_types_allowed = True
+
+    def configure_instance(self, runner: Runner) -> Dict:
+        """Configure instance."""
```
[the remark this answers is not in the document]
Abubakarr99: I did it another way.

tests/unit/backend/test_aws.py
```diff
@@ -0,0 +1,76 @@
...
+    )
+    return runner_group
+
+
+@fixture()
+def aws_runner(runner: Runner, aws_group: RunnerGroup) -> Runner:
+    # Cleanup and return a runner for testing
+    aws_group.backend.delete(runner)
+    return runner
+
```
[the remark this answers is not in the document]
Abubakarr99: ok

runner_manager/backend/aws.py
```diff
@@ -16,66 +18,130 @@ class AWSBackend(BaseBackend):
...
         """Get a runner."""
+        instance = {}
         try:
             instance = self.client.describe_instances(InstanceIds=[instance_id])
-            instance_id = instance["Instances"][0]["InstanceId"]
         except ClientError as e:
-            if e.response["Error"]["Code"] == "InvalidInstanceID.NotFound":
+            error = e.response.get("Error", {})
+            if error.get("Code") == "InvalidInstanceID.NotFound":
                 log.error(f"Instance {instance_id} not found.")
```
nootal, line 67: It's weird to continue the method after this. If we got not instance, we should return right here.

runner_manager/backend/aws.py
```diff
@@ -16,66 +18,130 @@ class AWSBackend(BaseBackend):
...
+        try:
+            reservations = instance.get("Reservations", [])
+            if reservations:
+                instances = reservations[0].get("Instances", [])
+                if instances:
+                    instance_id = instances[0]["InstanceId"]
+                else:
+                    raise ValueError(f"No instances found for {instance_id}")
+            else:
+                raise ValueError(f"No reservations found for {instance_id}")
```
nootal, line 79: See comment above, but that makes no sense to get here if the instance was simply not found

runner_manager/backend/aws.py
```diff
@@ -16,66 +18,130 @@ class AWSBackend(BaseBackend):
...
-            if e.response["Error"]["Code"] == "InvalidInstanceID.NotFound":
+            error = e.response.get("Error", {})
+            if error.get("Code") == "InvalidInstanceID.NotFound":
                 log.error(f"Instance {instance_id} not found.")
             else:
                 raise e
+        try:
+            reservations = instance.get("Reservations", [])
+            if reservations:
+                instances = reservations[0].get("Instances", [])
```
nootal, line 73: Same here

runner_manager/backend/aws.py
```diff
@@ -16,66 +18,130 @@ class AWSBackend(BaseBackend):
...
+                        instance_id=instance.get("InstanceId", ""),
+                        busy=False,
+                    )
+                    runners.append(runner)
+        return runners or []
+
+    def update(self, runner: Runner) -> Runner:
+        """Update a runner."""
+        try:
+            if runner.instance_id:
```
nootal, line 122: Don't put a "if" in a try statement in this case. You want the content of the "try" to be as small as possible

nootal (changes requested):

runner_manager/backend/aws.py
Abubakarr99: yeah I corrected that

runner_manager/backend/aws.py
Abubakarr99: ok correcting it

runner_manager/backend/aws.py
```diff
@@ -16,66 +18,130 @@ class AWSBackend(BaseBackend):
...
+                try:
+                    runners.append(
+                        Runner.find(
+                            Runner.instance_id == instance.get("InstanceId", "")
+                        ).first()
+                    )
+                except NotFoundError:
+                    runner = Runner(
+                        name=instance.get("InstanceId", ""),
+                        instance_id=instance.get("InstanceId", ""),
```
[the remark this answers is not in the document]
Abubakarr99: ok got it. changing it now

runner_manager/backend/aws.py
Abubakarr99: ok noted

runner_manager/backend/aws.py
```diff
@@ -16,66 +18,130 @@ class AWSBackend(BaseBackend):
...
+                    InstanceIds=[runner.instance_id],
+                    Filters=[
+                        {"Name": "instance-state-name", "Values": ["running"]},
+                        {"Name": "tag-key", "Values": ["runner-manager"]},
+                    ],
+                )
+        except ClientError as e:
+            error = e.response.get("Error", {})
+            if error.get("Code") == "InvalidInstanceID.NotFound":
+                log.error(f"Instance {runner.instance_id} not found.")
```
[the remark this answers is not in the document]
Abubakarr99: yeah correct.

runner_manager/backend/aws.py
```diff
@@ -65,20 +64,17 @@ def get(self, instance_id: str) -> Runner:
             error = e.response.get("Error", {})
             if error.get("Code") == "InvalidInstanceID.NotFound":
                 log.error(f"Instance {instance_id} not found.")
+                return Runner.find(Runner.instance_id == instance_id).first()
```
nootal, line 67: I don't understand this part. If we don't find it, we still return it? What do we return here exactly?

runner_manager/backend/aws.py
```diff
@@ -96,52 +92,56 @@ def list(self) -> List[Runner]:
...
-        try:
-            if runner.labels and runner.instance_id:
+            except ClientError as e:
+                error = e.response.get("Error", {})
+                if error.get("Code") == "InvalidInstanceID.NotFound":
+                    log.error(f"Instance {runner.instance_id} not found.")
+                    return super().update(runner)
+                else:
+                    raise e
+        if runner.labels and runner.instance_id:
```
nootal, line 135: This could be nested in the already existing `if runner.instance_id` statement from above:

```suggestion
            if runner.labels:
```

runner_manager/backend/aws.py
```diff
@@ -96,52 +92,56 @@ def list(self) -> List[Runner]:
...
-                log.error(f"Instance {runner.instance_id} not found.")
-            else:
-                raise e
-        try:
-            if runner.labels and runner.instance_id:
+            except ClientError as e:
+                error = e.response.get("Error", {})
+                if error.get("Code") == "InvalidInstanceID.NotFound":
+                    log.error(f"Instance {runner.instance_id} not found.")
+                    return super().update(runner)
```
nootal, line 132: Same question: we don't find it, so we return it?

nootal (changes requested):

runner_manager/backend/aws.py
Abubakarr99: so it tries to find the instance and return none if no match is found.

nootal: Why not returning `None` then? Why this call to `Runner.find`?

runner_manager/backend/aws.py
```diff
@@ -27,37 +27,32 @@ def create(self, runner: Runner) -> Runner:
...
-                log.error(f"Instance {runner.instance_id} malformed.")
-                pass
-            else:
-                raise e
+            except ClientError as e:
+                error = e.response.get("Error", {})
+                if error.get("Code") == "InvalidInstanceID.NotFound":
+                    log.error(f"Instance {runner.instance_id} not found.")
+                elif error.get("Code") == "InvalidInstanceID.Malformed":
+                    log.error(f"Instance {runner.instance_id} malformed.")
```
tcarmet, line 53: Ensure the delete function from the parent gets called otherwise we face the same issue as with the gcloud backend where it would be stuck in a infinite loop.
```suggestion
                if error.get("Code") == "InvalidInstanceID.NotFound":
                    log.error(f"Instance {runner.instance_id} not found.")
                    return super().delete(runner)
                elif error.get("Code") == "InvalidInstanceID.Malformed":
                    log.error(f"Instance {runner.instance_id} malformed.")
                    return super().delete(runner)
```

runner_manager/backend/aws.py
tcarmet: Something is up indeed, I think this function should retrieve the instance name thanks to the instance_id. And then try to find the runner with the `name` instead of the `instance_id` otherwise there's no point of using it.

tcarmet (changes requested):

runner_manager/backend/aws.py
tcarmet: We could find an base exception that can be raised or create a new one under `runner_manager/exceptions.py`


But for now let's leave as is and raise NotFoundError just for consistency with the base backend. this method is not yet used in the rest of the code and might be deleted at the end.

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,147 @@
...
+                raise e
+        try:
+            reservations = instance.get("Reservations")
+            if reservations:
+                instances = reservations[0].get("Instances")
+                if instances:
+                    instance_id = instances[0]["InstanceId"]
+        except Exception:
+            raise
+        return Runner.find(Runner.instance_id == instance_id).first()
```
tcarmet, line 78:
```suggestion
        return Runner.find((Runner.name == name) | (Runner.instance_id == instance_id)).first()
```

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,147 @@
...
+                        {"Name": "tag-key", "Values": ["runner-manager"]},
+                    ],
+                )
+            except ClientError as e:
+                error = e.response.get("Error", {})
+                if error.get("Code") == "InvalidInstanceID.NotFound":
+                    log.error(f"Instance {runner.instance_id} not found.")
+                    return super().update(runner)
+                else:
+                    raise e
```
tcarmet, line 134:
```suggestion
```

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,147 @@
...
+                    self.client.create_tags(
+                        Resources=[runner.instance_id],
+                        Tags=[
+                            {"Key": label.name, "Value": label.name}
+                            for label in runner.labels
+                        ],
+                    )
+                except Exception as e:
+                    log.error(e)
+                    raise e
```
tcarmet, line 146:
```suggestion

            try:
                self.client.create_tags(
                    Resources=[runner.instance_id],
                    Tags=[{"Key": "tag:status", "Value": runner.status}]
                )
            except Exception as e:
                log.error(e)
                raise e
```

runner_manager/models/backend.py
```diff
@@ -116,3 +116,60 @@ def configure_instance(self, runner: Runner) -> Instance:
...
+                }
+            ],
+            "UserData": self.user_data,
+            "MaxCount": self.max_count,
+            "MinCount": self.min_count,
+            "BlockDeviceMappings": self.block_device_mappings,
+        }
+        if self.tags:
+            instance_config["TagSpecifications"][0]["Tags"].extend(
+                [{"Key": key, "Value": value} for key, value in self.tags.items()]
```
tcarmet, line 173: prefix key by `f"tag:{key}"`

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,147 @@
...
+                error = e.response.get("Error", {})
+                if error.get("Code") == "InvalidInstanceID.NotFound":
+                    log.error(f"Instance {runner.instance_id} not found.")
+                elif error.get("Code") == "InvalidInstanceID.Malformed":
+                    log.error(f"Instance {runner.instance_id} malformed.")
+                else:
+                    raise e
+        return super().delete(runner)
+
+    def get(self, instance_id: str) -> Runner:
```
tcarmet, line 58: let's remove it.

Abubakarr99: ok

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,109 @@
...
+            try:
+                self.client.terminate_instances(InstanceIds=[runner.instance_id])
+            except ClientError as e:
+                error = e.response.get("Error", {})
+                if error.get("Code") == "InvalidInstanceID.NotFound":
+                    log.error(f"Instance {runner.instance_id} not found.")
+                    return super().delete(runner)
+                elif error.get("Code") == "InvalidInstanceID.Malformed":
+                    log.error(f"Instance {runner.instance_id} malformed.")
+                    return super().delete(runner)
```
nootal, line 55:
```suggestion
                elif error.get("Code") == "InvalidInstanceID.Malformed":
                    log.error(f"Instance {runner.instance_id} malformed.")
```
Better to just pass here so you'll catch up with the final `return super().delete(runner)`; no need to repeat it

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,109 @@
...
+                else:
+                    raise e
+        return super().delete(runner)
+
+    def list(self) -> List[Runner]:
+        """List runners."""
+        runners: List[Runner] = []
```
nootal, line 62: minor, but we should not define it here if we don't use it in the next statement; this could be defined once we are sure we have a `reservations` object

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,109 @@
...
+                            Runner.instance_id == instance_id,
+                        ).first()
+                    )
+                except NotFoundError:
+                    runner = Runner(
+                        name=instance_id,
+                        instance_id=instance_id,
+                        busy=False,
+                    )
+                    runners.append(runner)
```
nootal, line 94:
```suggestion
                try:
                    runner = Runner.find(
                        Runner.instance_id == instance_id,
                    ).first()
                except NotFoundError:
                    runner = Runner(
                        name=instance_id,
                        instance_id=instance_id,
                        busy=False,
                    )
                runners.append(runner)
```
Better this way to understand the goal is to find a runner or create it then we do the same thing, we add it to the list.

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,109 @@
...
+        return runners or []
```
nootal, line 95:
```suggestion
        return runners
```
As `runners` is already defined as an empty list, no need to this

runner_manager/backend/aws.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def update(self, runner: Runner) -> Runner:
+        """Update a runner."""
+        if runner.instance_id:
+
```
nootal, line 100:
```suggestion
```

runner_manager/models/backend.py
```diff
@@ -116,3 +116,60 @@ def configure_instance(self, runner: Runner) -> Instance:
...
+    """Configuration for AWS backend."""
+
+    region: str = "us-west-2"
+    subnet_id: str
+
+
+class AWSInstanceConfig(InstanceConfig):
+    """Configuration for AWS backend instance."""
+
+    image: str = "ami-0735c191cf914754d"  # Ubuntu 22.04 for us-west-2
```
nootal, line 131: Need to make sure this is a gp3 AMI for the disks

runner_manager/models/backend.py
```diff
@@ -116,3 +116,60 @@ def configure_instance(self, runner: Runner) -> Instance:
...
+            ],
+            "UserData": self.user_data,
+            "MaxCount": self.max_count,
+            "MinCount": self.min_count,
+            "BlockDeviceMappings": self.block_device_mappings,
+        }
+        if self.tags:
+            instance_config["TagSpecifications"][0]["Tags"].extend(
+                [{"Key": key, "Value": value} for key, value in self.tags.items()]
+            )
```
nootal, line 174: I would find it easier to process (and to read) if you were defining the tags before the whole dict, like:

```suggestion
        tags = {
            "Name": runner.name,
            "runner-manager": runner.manager,
        }
        tags.update(self.tags)
        instance_config = {
            "ImageId": self.image,
            "InstanceType": self.instance_type,
            "SubnetId": self.subnet_id,
            "SecurityGroupIds": self.security_group_ids,
            "TagSpecifications": [
                {
                    "ResourceType": "instance",
                    "Tags": [{"Key": key, "Value": value} for key, value in tags.items()]
                }
            ],
            "UserData": self.user_data,
            "MaxCount": self.max_count,
            "MinCount": self.min_count,
            "BlockDeviceMappings": self.block_device_mappings,
        }
```

nootal (changes requested):

runner_manager/backend/aws.py
Abubakarr99: ok

runner_manager/backend/aws.py
tcarmet: Problem was not addressed. We need to ensure that the runner was deleted on the database side on some situations.

tcarmet (changes requested):
There's one conversation that was solved without a fix preventing me from approving this PR. Otherwise LGTM

runner_manager/models/backend.py
Abubakarr99: I think its not tied to the ami you can set that while creating the volume

runner_manager/backend/aws.py
Abubakarr99: removed the get method.

runner_manager/backend/aws.py
nootal: Same for all above `return`

runner_manager/models/backend.py
nootal: OK nice

runner_manager/models/backend.py
```diff
@@ -116,3 +116,62 @@ def configure_instance(self, runner: Runner) -> Instance:
...
+        }
+        block_device_mappings = [
+            {
+                "DeviceName": "/dev/sda1",
+                "Ebs": {
+                    "VolumeType": self.volume_type,
+                },
+            }
+        ]
+        tags.update(self.tags)
```
nootal, line 158:
```suggestion
        }
        tags.update(self.tags)
        block_device_mappings = [
            {
                "DeviceName": "/dev/sda1",
                "Ebs": {
                    "VolumeType": self.volume_type,
                },
            }
        ]
```

nootal (changes requested):

tcarmet (approved):

nootal (approved):

PR #362, opened 2023-08-21
PTFE-753 attach gcp runners to github

tcarmet:


tests/unit/conftest.py
```diff
@@ -49,7 +49,7 @@ def runner(settings) -> Runner:
         runner_group_id=1,
         status="online",
         busy=False,
-        labels=[],
+        labels=[{"name": "label"}],
```
Abubakarr99, line 52:
```suggestion
        labels=RunnerLabel([{"name": "label"}]),
```

Abubakarr99 (commented):
LGTM tiny suggestion for labels

tests/unit/conftest.py
tcarmet:
```suggestion
        labels=[RunnerLabel(name="label")],
```

PR #366, opened 2023-08-22
PTFE-683 implement reset endpoint

tcarmet:


runner_manager/jobs/reset.py
```diff
@@ -0,0 +1,24 @@
...
+def group(pk: str):
+    """Job to reset a runner group.
+
+    Args:
+        pk (str): Runner group primary key
+    """
+    github: GitHub = get_github()
+    try:
+        group = RunnerGroup.get(pk)
+        group.reset(github)
```
nootal, line 21: Same comment as in https://github.com/scality/runner-manager/pull/365 regarding the try statement containing lines that are not catchable by the except, so they should be put outside of it

runner_manager/jobs/reset.py
```diff
@@ -0,0 +1,24 @@
...
+    except NotFoundError as e:
+        log.error(f"Runner group {pk} not found")
+        raise e
```
nootal, line 24:
```suggestion
    except NotFoundError:
        log.error(f"Runner group {pk} not found")
        raise
```
If we don't care about the content of the exception, we should not name it

runner_manager/models/runner_group.py
```diff
@@ -256,6 +256,25 @@ def healthcheck(
...
+                    ).parsed_data
+                )
+                runner = runner.update_status(github_runner)
+            if runner.status == RunnerStatus.offline:
+                self.delete_runner(runner)
+                token = github.rest.actions.create_registration_token_for_org(
+                    org=self.organization
+                )
+                self.create_runner(token.parsed_data)
+        return self
```
nootal, line 276: Same comment as above. We already have "self" somewhere else because we call the method on this object itself. What is the point of returning it.

nootal (changes requested):

runner_manager/models/runner_group.py
```diff
@@ -256,6 +256,25 @@ def healthcheck(
...
+        """Reset runner group."""
+        runners = self.get_runners()
+        for runner in runners:
+            if runner.id is not None:
+                github_runner: GitHubRunner = (
+                    github.rest.actions.get_self_hosted_runner_for_org(
+                        self.organization, runner.id
+                    ).parsed_data
+                )
+                runner = runner.update_status(github_runner)
```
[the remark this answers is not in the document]
tcarmet: indeed mistake on my end.
```suggestion
                runner.update_status(github_runner)
```

runner_manager/models/runner_group.py
```diff
@@ -256,6 +256,25 @@ def healthcheck(
...
+            if runner.status == RunnerStatus.offline:
```
tcarmet, line 270:
```suggestion
            if not runner.is_active 
```

runner_manager/models/runner_group.py
```diff
@@ -268,6 +268,24 @@ def healthcheck(
...
 
+    def reset(self, github: GitHub):
+        """Reset runner group."""
+        for runner in self.get_runners():
+            if runner.id is not None:
+                (
+                    github.rest.actions.get_self_hosted_runner_for_org(
+                        self.organization, runner.id
+                    ).parsed_data
+                )
```
tcarmet, line 279: We recently made some change into this function and now `update_from_github` does this API call.

```suggestion
```

runner_manager/models/runner_group.py
```diff
@@ -268,6 +268,24 @@ def healthcheck(
...
+                runner.update_from_github(github)
+            if not runner.is_active:
+                self.delete_runner(runner)
+                github.rest.actions.create_registration_token_for_org(
+                    org=self.organization
+                )
```
tcarmet, line 285: Another change we introduced and this request is now done by `create_runner`
```suggestion
```

tcarmet (commented):
Thanks for taking over @MartinS926

With the two suggestions in, I believe we are good to go.

runner_manager/models/runner_group.py
nootal: Yeah and I don't really understand why those parenthesis around it.

runner_manager/models/runner_group.py
```diff
@@ -268,6 +268,24 @@ def healthcheck(
...
+        """Reset runner group."""
+        for runner in self.get_runners():
+            if runner.id is not None:
+                (
+                    github.rest.actions.get_self_hosted_runner_for_org(
+                        self.organization, runner.id
+                    ).parsed_data
+                )
+                runner.update_from_github(github)
+            if not runner.is_active:
```
nootal, line 281: So if the runner is active we do nothing?

MartinS926: if the runner is active it is not a good choice to reset it

nootal (approved):

PR #376, opened 2023-08-24
healthcheck runner checkup

tcarmet:
The healthcheck function would check if runner.id is defined in order to retrieve its state from github.

The problem with that is that unless the runner has started a job, we would never retrieve its id, so idle runner would never be able to get an updated status and be forever considered as offline by the runner manager.

To fix this issue, we are using another endpoint to create the registration token for runners, which allow us to specify the full runner configuration as well as retrieve the id that will be given.
Thanks to this change we will always have an `runner.id` to work with.

### DIFF
--- a/images/runner/entrypoint.sh
+++ b/images/runner/entrypoint.sh
@@ -2,20 +2,9 @@
 
 echo "Starting runner..."
 echo "RUNNER_ORG: ${RUNNER_ORG}"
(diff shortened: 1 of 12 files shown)


tcarmet:
I believe there's still an issue with the startup.sh script:

- [ ] Debug startup.sh and fix issue

nootal (approved):

PR #380, opened 2023-08-25
PTFE-780 Fix bug with runner status and API status update

tcarmet:


runner_manager/models/runner.py
```diff
@@ -153,9 +158,9 @@ def update_from_github(self, github: GitHub) -> "Runner":
                     org=self.organization, runner_id=self.id
                 ).parsed_data
             )
-            self.status = RunnerStatus(self.status)
+            self.status = RunnerStatus(github_runner.status)
```
tcarmet, line 161: This is why we were always considering the runner as "offline".

runner_manager/models/runner.py
```diff
@@ -25,7 +26,6 @@
 
 class RunnerStatus(str, Enum):
     online = "online"
-    idle = "idle"
     offline = "offline"
```
tcarmet, line 29: For the record I double checked your statement regarding status @Abubakarr99, and indeed discovered that there's no such thing as a `idle` status, only online and offline, then a runner is considered as online or active with the `busy` flag. Made the according changes.

tests/unit/jobs/test_healthchecks.py
```diff
@@ -91,23 +101,16 @@ def test_time_to_start(runner: Runner, settings: Settings):
...
 
-def test_need_new_runner(runner_group: RunnerGroup, github: GitHub):
-    runner_group.max = 2
-    runner_group.min = 1
-    runner_group.save()
-    assert runner_group.need_new_runner is True
-    runner_group.create_runner(github)
-    assert runner_group.need_new_runner is False
-
-
```
tcarmet, line 110: Duplicated test, removing it.

PR #381, opened 2023-08-28
PTFE-791: queue runners

Abubakarr99:
This would allow us to queue runners when we reach the limit. The queue is decremented if a runner is created and incremented if a runner could not be created

### DIFF
--- a/runner_manager/jobs/workflow_job.py
+++ b/runner_manager/jobs/workflow_job.py
@@ -21,12 +21,12 @@ def completed(webhook: WorkflowJobCompleted) -> int:
         log.warning(f"Runner {webhook.workflow_job.runner_name} not found")
         return 0
     log.info(f"Deleting runner {runner.name} in group {runner_group.name}")
-    if runner_group.queued > 0:
-        log.info(f"Runner group {runner_group.name} has a queue")
-        log.info(f"Creating new runner for group {runner_group.name}")
+    delete = runner_group.delete_runner(runner)
+    if runner_group.need_new_runner:
+        log.info(f"Runner group {runner_group.name} needs a new runner")
         github: GitHub = get_github()
         runner_group.create_runner(github)
-    return runner_group.delete_runner(runner)
+    return delete
 
 
 def in_progress(webhook: WorkflowJobInProgress) -> str | None:
--- a/runner_manager/jobs/workflow_job.py
+++ b/runner_manager/jobs/workflow_job.py
@@ -20,6 +20,12 @@ def completed(webhook: WorkflowJobCompleted) -> int:
     if not runner_group or not runner:
         log.warning(f"Runner {webhook.workflow_job.runner_name} not found")
         return 0
+    log.info(f"Deleting runner {runner.name} in group {runner_group.name}")
+    if runner_group.queued > 0:
+        log.info(f"Runner group {runner_group.name} has a queue")
+        log.info(f"Creating new runner for group {runner_group.name}")
+        github: GitHub = get_github()
+        runner_group.create_runner(github)
     return runner_group.delete_runner(runner)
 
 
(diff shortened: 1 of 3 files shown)


runner_manager/models/runner_group.py
```diff
@@ -162,7 +166,7 @@ def need_new_runner(self) -> bool:
         runners = self.get_runners()
         not_active = len([runner for runner in runners if runner.is_active is False])
         count = len(runners)
-        return not_active < self.min and count < self.max
+        return (not_active < self.min and count < self.max) or self.queued > 0
```
tcarmet, line 169: if we reach max but we have a queued runner we don't want to return `True` but `False`.

```suggestion
        return (not_active < self.min or self.queued > 0) and count < self.max
```

Not so sure about my suggestion so please create a testing scenario that has queued runners and max is == count

tcarmet (changes requested):

runner_manager/jobs/workflow_job.py
```diff
@@ -20,6 +20,12 @@ def completed(webhook: WorkflowJobCompleted) -> int:
     if not runner_group or not runner:
         log.warning(f"Runner {webhook.workflow_job.runner_name} not found")
         return 0
+    log.info(f"Deleting runner {runner.name} in group {runner_group.name}")
+    if runner_group.queued > 0:
```
tcarmet, line 24:
```suggestion
    if runner_group.need_new_runner:
```

runner_manager/jobs/workflow_job.py
```diff
@@ -20,6 +20,12 @@ def completed(webhook: WorkflowJobCompleted) -> int:
...
+        log.info(f"Runner group {runner_group.name} has a queue")
+        log.info(f"Creating new runner for group {runner_group.name}")
+        github: GitHub = get_github()
+        runner_group.create_runner(github)
     return runner_group.delete_runner(runner)
```
tcarmet, line 29:
```suggestion
    log.info(f"Deleting runner {runner.name} in group {runner_group.name}")
    delete = runner_group.delete_runner(runner)
    if runner_group.need_new_runner:
        log.info(f"Runner group {runner_group.name} needs a new runner")
        github: GitHub = get_github()
        runner_group.create_runner(github)
    return delete
```

tests/unit/models/test_runner_group.py
```diff
@@ -149,9 +149,15 @@ def test_need_new_runner(runner_group: RunnerGroup, github: GitHub):
...
     runner.status = RunnerStatus.online
     runner.busy = True
     runner.save()
     Migrator().run()
     assert runner_group.need_new_runner is True
+    # increase the queue by simulating a new job
+    runner_group.queued += 1
+    assert runner_group.need_new_runner is True
+    runner = runner_group.create_runner(github)
+    assert runner_group.queued == 0
```
tcarmet, line 163: New additional test case with more attention to the queue itself.

runner_manager/models/runner_group.py
```diff
@@ -129,9 +130,12 @@ def create_runner(self, github: GitHub) -> Runner | None:
...
             runner.save()
             runner.generate_jit_config(github)
-
+            if self.queued > 0:
+                self.queued -= 1
             return self.backend.create(runner)
-        return None
+        else:
+            self.queued += 1
+            return None
```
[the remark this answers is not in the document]
nootal: Can it be negative? Or else is =0?
And don't need to return None, just return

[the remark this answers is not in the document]
tcarmet: else is we couldn't created a runner because we reached the maximum allowed so we're adding it up.

There's a part of the code missing to review this part accordingly due to the diff size:
```python
if count < self.max and self.id
```

PR #383, opened 2023-08-28
Fix update status when runner is not found

tcarmet:


runner_manager/models/runner.py
```diff
@@ -151,16 +152,23 @@ def time_to_start_expired(self, timeout: timedelta) -> bool:
...
+                    github.rest.actions.get_self_hosted_runner_for_org(
+                        org=self.organization, runner_id=self.id, headers=headers
+                    ).parsed_data
+                )
+                self.status = RunnerStatus(github_runner.status)
+                self.busy = github_runner.busy
+            except RequestFailed:
+                log.info(f"Runner {self.name} does not exist anymore.")
+                self.status = RunnerStatus.offline
+                self.busy = False
```
nootal, line 170: As we use a specific Exception, I guess it can happen only in the first call of the try statement, so I would prefer having this instead:

```suggestion
            except RequestFailed:
                log.info(f"Runner {self.name} does not exist anymore.")
                self.status = RunnerStatus.offline
                self.busy = False
            else:
                self.status = RunnerStatus(github_runner.status)
                self.busy = github_runner.busy
```

That's more pythonic.

nootal (approved):
OK but I suggested a change anyway.

runner_manager/models/runner.py
tcarmet: Didn't knew we could do that, I like it.

PR #392, opened 2023-08-30
PTFE-799: aws runner manager startup

Abubakarr99:


runner_manager/models/backend.py
```diff
@@ -139,17 +139,27 @@ class AWSInstanceConfig(InstanceConfig):
...
+    startup_script: str = startup_sh.as_posix()
 
     class Config:
         arbitrary_types_allowed = True
 
+    def runner_env(self, runner: Runner) -> str:
+        env_vars = ""
+        env: RunnerEnv = super().runner_env(runner)
+        for key, value in env.dict().items():
+            env_vars += f"{key}={value} "
```
tcarmet, line 151:
```suggestion
            env_vars += f"{key}={value}"
```

tcarmet (commented):
* Set RunnerEnv as tags in the EC2 instance.
* In startup.sh follow the same method used to retrieve metadata from GCP, but with AWS.
* Run the script with no modification.
* optionally  add your ssh key in the config for debugging, hardcode during development if needed.

tcarmet (approved):

PR #418, opened 2023-09-06
PTFE-825 Add create_runner method for runner_group

MartinS926:


runner_manager/jobs/runner.py
```diff
@@ -0,0 +1,9 @@
+from runner_manager import RunnerGroup
+from runner_manager.clients.github import GitHub
+from runner_manager.dependencies import get_github
+
+
+def create_runner(group: RunnerGroup) -> str | None:
+    github: GitHub = get_github()
+    runner = group.create_runner(github)
+    return runner.pk if runner is not None else runner
```
nootal, line 9: `else runner` means returning None. We can write better then.

```suggestion
    if runner is not None:
        return runner.pk
```

nootal (changes requested):

tcarmet (changes requested):
Overall it's looking good, missing a couple of tweaks but most importantly I see no test made for the method introduced.

Checkout how tests are made into this [file](https://github.com/scality/runner-manager/blob/main/tests/api/test_runner_groups_router.py) and add some new test cases into it.

runner_manager/jobs/create.py
```diff
@@ -0,0 +1,19 @@
...
+def runner(group: RunnerGroup) -> str | None:
+    """Job to create a runner within a group.
+
+    Args:
+        RunnerGroup: Runner group to create a runner in
+    """
+    github: GitHub = get_github()
+    runner = group.create_runner(github)
+    if runner is not None:
+        return runner.pk
```
MartinS926, line 19: if runer is None, it means that the program couldnt create the runner ? What do we do in this case ?

tcarmet: Correct, each runner group has a max number of runner allowed, a case where runner is None would be that we asked a runner to be created but we reached `max`.

So in this case we do nothing, as you asked the runner to be created it has been counted in the queue and will be created later on by the program. It's all good here.

runner_manager/jobs/create.py
tcarmet, line 1: Rename file to `runner.py`

tcarmet (approved):
small comments to address but LGTM

PR #438, opened 2023-09-12
PTFE-782 Add synchronisation method

MartinS926:


runner_manager/jobs/startup.py
```diff
@@ -17,19 +17,28 @@
 log = logging.getLogger(__name__)
 
 
-def create_runner_groups(settings: Settings, github: GitHub):
+def sync_runner_groups(
+    settings: Settings, github: GitHub
+):  # creation des runner groups
```
tcarmet, line 22:
```suggestion
):
```

runner_manager/jobs/startup.py
```diff
@@ -17,19 +17,28 @@
...
+            runner_group.update(**runner_group_config.dict())
             runner_group.save(github=github)
+            existing_groups.remove(runner_group)
         else:
-            runner_group: RunnerGroup = RunnerGroup.find_from_base(runner_group_config)
+            runner_group: RunnerGroup = RunnerGroup(**runner_group_config.dict())
+            runner_group.save(github=github)
 
-            existing_groups.remove(runner_group)
-            runner_group.update(**runner_group_config.dict())
```
tcarmet, line 32: Review this whole block here, I don't think we should see any diff considering our implementation.

tcarmet (changes requested):

runner_manager/models/runner_group.py
```diff
@@ -200,6 +214,10 @@ def need_new_runner(self) -> bool:
     def create_github_group(self, github: GitHub) -> GitHubRunnerGroup:
         """Create a GitHub runner group."""
 
+        exists = self.github_group_exists(github)
+        if exists is not None and self.id is None:
+            self.id = exists.id
+
```
tcarmet, line 220: Please forgive me for my previous suggestion, I think I was tired, this is more towards what I was trying to achieve:
```suggestion
        if self.id is None:
            exists = self.github_group_exists(github)
            self.id = exists.id if exists is not None else None

```

There's most likely a prettier/better way to do it, all I'm trying to achieve here is to not look for the group if we already have `self.id` defined.

runner_manager/jobs/startup.py
tcarmet: One example I can find is line 31: https://github.com/scality/runner-manager/pull/438/files#diff-2c00abe7ba70ca4a271cec7b4edac8ff8fe9475fd6a0956f556db85b8b569a31R31

I think the condition should be `if runner_group_config.name ...` instead of `if runner_group_config ...`

tcarmet (approved):

PR #491, opened 2023-12-04
PTFE-1196 reduce impact of missed webhooks

tcarmet:


tests/strategies.py
```diff
@@ -127,7 +128,7 @@ class Repo(Repository):
     github_base_url=st.just("http://localhost:4010"),
     github_token=st.just("test"),
     time_to_live=st.integers(1, 60),
-    timeout_runner=st.integers(1, 10),
+    timeout_runner=st.integers(120, 600),
```
nootal, line 131: What are those exactly?

PR #503, opened 2023-12-15
PTFE-1196 ping redis on healthcheck calls

tcarmet:


nootal (approved):
I'm all for pinging Redis, but is this going to fix the issue where the API goes south and returns "Internal Server Error" when redis is back on track?

tcarmet:
> I'm all for pinging Redis, but is this going to fix the issue where the API goes south and returns "Internal Server Error" when redis is back on track?

The internal server errors were returned because the API was unable to contact the redis server. The reason why it was unable to do so remains a mistery, but regardless of the reason the healthcheck will fail (especially as there could be multiple reason that lead to the same symptom). 

When a healthcheck fails, [Kubernetes will restart](https://github.com/scality/runner-manager/blob/main/manifests/base/runner-manager/main/deployment.yaml#L37-L46) the service accordingly.

Now, it is possible that a ping to redis may succeed but other request won't, but before we go into the route of handling all possibilities, let's start with a simple one first.

nootal (approved):

PR #534, opened 2024-02-09
PTFE-1418 | PTFE-1320 start sanity check before job execution

tcarmet:
Make use of hooks to be executed on the runner before and after a job starts.

An extra step will appear in the job logs, displaying the logs of the hook while being executed.
Screenshot 2024-02-09 at 4 08 21 PM


nootal (commented):
no tested but I did not see anything weird

Issue #567, opened 2024-04-03
docs: mermaid diagrams don't render


lewismiddleton:
**Expected Behaviour**
They should render out as per the [mkdocs-material](https://squidfunk.github.io/mkdocs-material/reference/diagrams/?h=mermaid) documentation.

**Actual Behaviour**
The mermaid diagrams over at https://scality.github.io/runner-manager/development/workflows/ don't render. Instead they show raw code blocks. 

<img width="1344" alt="Screenshot 2024-04-03 at 1 01 11 am" src="https://github.com/scality/runner-manager/assets/66401219/8147f546-5df1-4d56-a8fe-153a6d1e52f1">

lewismiddleton:
I put two implementations up at #568 and #569, you can choose which you prefer or do some digging of your own.

I'd suggest choosing the `mermaid2` plugin implementation in #568 and let `poetry` manage the dependency. Compared to the hard coded mermaid.js version in the `extra_javascript` implementation that will probably not get updated.

On the other hand, [this comment](https://github.com/squidfunk/mkdocs-material/issues/693#issuecomment-743965446) from a `mkdocs-material` maintainer seems to suggest you should not use a plugin unless you need 'advanced behaviour'.

lewismiddleton:
I'll add some context here about a small rabbit hole about `<div>` vs `<pre><code>`

In the page build logs for `mkdocs` in #568, `mermaid2` is doing something with the `<pre><code>` blocks and turning them into `<div>` suggesting that the `mermaid2` project knows something about the incompatibility with `fence_code_format`.

Here's some logs
```
vscode ➜ /workspaces/runner-manager (main) $ poetry run mkdocs build --strict
INFO    -  MERMAID2  - Initialization arguments: {}
INFO    -  MERMAID2  - Using javascript library (10.4.0):
              https://unpkg.com/mermaid@10.4.0/dist/mermaid.esm.min.mjs
INFO    -  Cleaning site directory
INFO    -  Building documentation to directory: /workspaces/runner-manager/site
INFO    -  MERMAID2  - Page 'Workflows': found 2 diagrams (with <pre><code='[language-]mermaid'>), converting to <div>...
INFO    -  MERMAID2  - Page 'Workflows': found 2 diagrams, adding scripts
INFO    -  Documentation built in 0.90 seconds
```

[...]

PR #568, opened 2024-04-03
docs: Render mermaid graphs

lewismiddleton:
Add the plugin as a dependency as per [installation guide](https://github.com/fralau/mkdocs-mermaid2-plugin#installation).
`poetry add mkdocs-mermaid2-plugin --group docs`

Add to the `mkdocs.yml` via [plugins syntax](https://www.mkdocs.org/dev-guide/plugins/)

This on it's own is sufficient to get the diagrams to render.

Resolves: #567


lewismiddleton:
Adding a `fence_mermaid_custom` format to superfences as per [mermaid2 plugin Material theme config](https://github.com/fralau/mkdocs-mermaid2-plugin#additional-settings-for-the-material-theme) didn't seem to change behaviour but [seems like is best practice](https://github.com/fralau/mkdocs-mermaid2-plugin/issues/77#issuecomment-1642100370) if you want the diagrams to inherit styling from any theming you choose.

```
plugins: 
  - mermaid2

markdown_extensions:
  - pymdownx.superfences:
      custom_fences:
        - name: mermaid
          class: mermaid
          format: !!python/name:mermaid2.fence_mermaid_custom
```

mkdocs.yml
```diff
@@ -56,7 +59,11 @@ markdown_extensions:
       emoji_generator: !!python/name:materialx.emoji.to_svg
   - pymdownx.inlinehilite
   - pymdownx.snippets
-  - pymdownx.superfences
+  - pymdownx.superfences:
+      custom_fences:
+        - name: mermaid
+          class: mermaid
+          format: !!python/name:mermaid2.fence_mermaid_custom
```
tcarmet, line 66: Hey @lewismiddleton! Thanks for your contribution and for spotting this issue, very much appreciated! 

In my other mkdocs project I did the following:
```suggestion
  - pymdownx.superfences:
      custom_fences:
        - name: mermaid
          class: mermaid
          format: !!python/name:pymdownx.superfences.fence_code_format
```

For me it seems like a healthier change as it doesn't require the extra dependency to be installed, would you mind giving it a shot see if it renders as you would like compared with `mermaid2`.

Thanks in advance!

lewismiddleton: Yeah I agree, a better solution if it works. I wasn't able to get that working, give it a try on your environment and see what you think?

You're right though in that [the docs](https://squidfunk.github.io/mkdocs-material/reference/diagrams/#configuration) say it should be sufficient. I'm curious to know what the difference is between this project and others (if there is actually an issue - idk maybe my environment is just differnet)

tcarmet: I did try it and it worked.

Worth mentioning though that I was running my doc website through a `mkdocs serve` and reaching it on `localhost:8000` through my browser and not via the `mkdocs build`, it seems like you were using the build method and I don't know if that may cause some differences in the rendering.

Regardless I have another website with those options hosted on GitHub pages that is working fine as well for rendering the mermaid graphs.

Let's give it a go with only the suggestion above, merge the PR, then if it's doesn't work we'll review the addition of the plugin.

lewismiddleton: Yeah I was using `poetry run mkdocs build --strict` but can confirm `mkdocs serve` works on my machine too.

I've reverted the commits and pushed the suggested change.

tcarmet (approved):

mkdocs.yml
tcarmet: Awesome! Thank you for your patience and making the suggested changes ❤️ 

And good to know it doesn't behave the same with a `mkdocs build`!

PR #576, opened 2024-04-09
aws: add support for instance profiles

harryfinbow:
Currently there is no way of assigning IAM roles to runner instances at creation time. This PR adds an additional field in the AWS backend which creates EC2 instances with the specified `IamInstanceProfile`, configurable by the `iam_instance_profile_arn` field in the `runner-manager.yaml` config.

This implementation only accepts instances profiles by their ARN and not by their name to keep the implementation as simple as possible.

### DIFF
--- a/runner_manager/models/backend.py
+++ b/runner_manager/models/backend.py
@@ -152,7 +152,7 @@ class AWSInstanceConfig(InstanceConfig):
     tags: Dict[str, str] = {}
     volume_type: VolumeTypeType = "gp3"
     disk_size_gb: int = 20
-    iam_instance_profile_name: Optional[str] = ""
+    iam_instance_profile_arn: str = ""
 
     def configure_instance(self, runner: Runner) -> AwsInstance:
         """Configure instance."""
@@ -188,7 +188,7 @@ class AWSInstanceConfig(InstanceConfig):
             ),
         ]
         iam_instance_profile: IamInstanceProfileTypeDef = IamInstanceProfileTypeDef(
-            Name=self.iam_instance_profile_name
+            Arn=self.iam_instance_profile_arn
         )
         return AwsInstance(
             ImageId=self.image,
--- a/runner_manager/models/backend.py
+++ b/runner_manager/models/backend.py
@@ -7,9 +7,9 @@ from mypy_boto3_ec2.literals import InstanceTypeType, VolumeTypeType
 from mypy_boto3_ec2.type_defs import (
     BlockDeviceMappingTypeDef,
     EbsBlockDeviceTypeDef,
+    IamInstanceProfileTypeDef,
     TagSpecificationTypeDef,
     TagTypeDef,
-    IamInstanceProfileTypeDef,
 )
 from pydantic import BaseModel, BaseSettings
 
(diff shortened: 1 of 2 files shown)


runner_manager/models/backend.py
```diff
@@ -184,6 +187,8 @@ def configure_instance(self, runner: Runner) -> AwsInstance:
                 Tags=tags,
             ),
         ]
+        iam_instance_profile: IamInstanceProfileTypeDef = IamInstanceProfileTypeDef(Name=self.iam_instance_profile_name)
```
harryfinbow, line 190: Not sure if this needs to be typed as it can never not be `IamInstanceProfileTypeDef`

runner_manager/models/backend.py
```diff
@@ -150,6 +152,7 @@ class AWSInstanceConfig(InstanceConfig):
     tags: Dict[str, str] = {}
     volume_type: VolumeTypeType = "gp3"
     disk_size_gb: int = 20
+    iam_instance_profile_name: Optional[str] = ""
```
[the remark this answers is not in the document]
tcarmet: It depends on how you want to treat it with your code I believe, and if you want to give it a default value to keep `pyright` happy (I think, not exactly sure). But yeah good call on following user_data for consistency.

runner_manager/models/backend.py
```diff
@@ -184,6 +187,9 @@ def configure_instance(self, runner: Runner) -> AwsInstance:
                 Tags=tags,
             ),
         ]
+        iam_instance_profile: IamInstanceProfileTypeDef = IamInstanceProfileTypeDef(
+            Name=self.iam_instance_profile_name
+        )
```
tcarmet, line 192: What happens if `iam_instance_profile_name` is not defined and `iam_instance_profile` becomes: `IamInstanceProfileTypeDef(Name="")`, will the EC2 instance be created accordingly?

runner_manager/models/backend.py
tcarmet: Yeah up to you, I have no personal preference and as long as the editor understand I'm happy. (which I think it does if you remove the type).

runner_manager/models/backend.py
harryfinbow: Passing in an empty string (e.g. not setting the value in the config file) successfully creates an EC2 instance without any IAM role associated with it which is what I would expect to happen

runner_manager/models/backend.py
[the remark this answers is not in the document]
harryfinbow: Running `pyright` solved this question for me pretty quickly as it did not like me making it optional so have removed it :)

runner_manager/models/backend.py
harryfinbow: As far as I can tell :) I have also had a last minute change of heart and changed it from name to ARN as I think thats the better of the two.

harryfinbow:
> LGTM, I see some linting stuff to be fixed let me know if they need some attention on my end.

I think I have fixed the linting issues (I had forgotten to run `trunk check` before submitting this PR) so hopefully a re-run of the checks should be all good

PR #583, opened 2024-04-19
Improvements to list method on aws and gcp

tcarmet:


nootal (approved):
I'm unable to understand if it's going to be better or worst, but i'll approve :p

PR #601, opened 2024-05-15
webhook: fix non-matching webhook secrets not being rejected

harryfinbow:
Fixes an issue which allowed webhooks to be processed with an incorrect webhook secret. 

This adds a case to return a `401` when the `verify` function returns false. The `verify` function also always returns false as it is being passed in JSON from a `WorkflowJob` webhook object instead of the request body (which is how GitHub generates it's signitures) meaning the signatures will never match (see https://github.com/yanyongyu/githubkit?tab=readme-ov-file#webhook-verification).

### DIFF
--- a/runner_manager/routers/webhook.py
+++ b/runner_manager/routers/webhook.py
@@ -1,6 +1,6 @@
 from typing import Annotated, Set
 
-from fastapi import APIRouter, Depends, Header, HTTPException, Security
+from fastapi import APIRouter, Depends, Header, HTTPException, Security, Request
 from githubkit.versions.latest.models import WebhookPing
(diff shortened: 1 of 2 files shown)


harryfinbow:
I am not familiar with FastAPI at all so if there is a better way of getting the request body without having to add `async` that you know of I am happy to change it :)

tcarmet (approved):
Thanks for providing this very valuable fix!

I don't have enough knowledge as well into why we have to use `async` to make this code works, but given that you have provided the appropriate test case and it is covered, and I know that using `async` with FastAPI is usually the way to go so LGTM and thanks again ❤️ !

PR #606, opened 2024-05-21
PTFE-1738 introducing job that will warn about leaks

tcarmet:


runner_manager/jobs/leaks.py
```diff
@@ -0,0 +1,32 @@
...
+    try:
+        group: RunnerGroup = RunnerGroup.get(pk)
+        backend_runners = group.backend.list()
+        group_runners = group.get_runners()
+        if len(backend_runners) > len(group_runners):
+            log.warning(f"Runner group {pk} has leaks")
+            for runner in backend_runners:
+                if runner not in group_runners:
+                    log.warning(f"Runner {runner} could be considered as a leak")
+            return True
```
nootal, line 29: put less stuff in the `try` statement, to make sure your `except` is properly targeted

nootal, line 29: could be put in an `else` statement to be extra explicit

nootal (changes requested):

nootal (approved):

Issue #625, opened 2024-06-26
Add metrics endpoint


lewismiddleton:
I'd like to scrape metrics about the internal runner-manager state. Specifically have it expose how many runners it's tracking in each of its runner groups.

This is useful because we've sometimes observed that AWS EC2 instances may get 'orphaned' meaning the runner-manager service has lost track of it and forgets to delete it. Monitoring these metrics would allow us to detect these 'orphaned' instances by alerting on when the number of EC2 machines != the tracked number of runners.

I propose using [Prometheus](https://prometheus.io/) instrumentation via the [prometheus-client library](https://pypi.org/project/prometheus-client). This will introduce a `/metrics` endpoint that is used for machine readable metrics. 

Are there any objections to this approach?

tcarmet:
No objections, it would be a very much welcome contribution. 

Quick note about the orphan situation, we recently noticed that in our kubernetes manifests we made a mistake on the redis volume path which in result would not persist our data and create orphans every time redis would reboot. (if that's the method of deployment you are using, worth double checking).

lewismiddleton:
I tried an implementation in #628 but I don't think that's the right approach. To get prometheus metrics, I'm now looking at something like dumping the Redis data as JSON using `get_runners()` upon request to `/groups/{name}/list`.

Then writing a custom prometheus exporter to consume that JSON and parse it into prometheus metrics.

PR #628, opened 2024-07-08
superseded: add prometheus metrics

lewismiddleton:
Related to #625.

This implementation is not working at a production level as it does not interact properly with the Redis Queue (https://github.com/rq/rq).

All the function calls happen on the worker container which means when the api container serves prometheus metrics from the in memory objects, it has no data.

The worker container has an [entrypoint using the rq binary](https://github.com/scality/runner-manager/blob/main/manifests/base/runner-manager/worker/deployment.yaml#L16) which means I don't think I can also expose a prometheus server for serving metrics as I don't have a hook into the startup.

I think this is a dead end but I'm putting the code up here for posterity. I think I'll reapproach #625 from a different perspective. Some options that could be explored on this implementation are:

1. Write a job function that can be sent to the worker container. The function would fetch the metrics from the prometheus client and then pass them back to the api server. 
- this has the issue that putting a job on the queue means the request becomes asynchronous, whereas the HTTP request to `/metrics` is synchronous. To avoid busy waiting we could maybe do something like an asyncio.sleep?


lewismiddleton:
closing in favour of #630

PR #629, opened 2024-07-08
api: add endpoint to list runners

lewismiddleton:
Adds an endpoint at `/groups/{name}/list` to list the runners in each runner group .

Relates to #625


lewismiddleton:
This implementation causes the `encoded_jit_config` to be exported as JSON. This could be a security risk in that it may allow someone with access to the runner-manager endpoint to register a runner using the same config.

I did some testing locally and I wasn't able to register a new runner using the same `encoded_jit_config` but I don't have high confidence that will always be the case.

lewismiddleton:
Trying to remove the `encoded_jit_config` from the output in [df057a2](https://github.com/lewismiddleton/runner-manager/commit/df057a276463255a3e47620cacdba30dfb694d8b) worked in the test suite but still outputted `jit_config` when deployed in a live environment. Unsure why that's the case.

tcarmet (approved):

PR #630, opened 2024-07-09
api: add metrics endpoint

### LINKED ISSUE #625: Add metrics endpoint

lewismiddleton:
Adds a metrics endpoint that computes updated metric values on each request.

On each request we:
- fetch runner groups from Redis
- get all runners for each group
- update a Gauge for the runners_count on each group
- render out the prometheus metrics in standard format

Resolves #625


tcarmet (approved):
LGTM 👌  Thanks for this contribution! I look forward to see it live.

lewismiddleton:
Ah I got a [lint error](https://github.com/scality/runner-manager/actions/runs/9861088056/job/27293380121#step:6:406), I've just [pushed the auto-fix](https://github.com/scality/runner-manager/compare/0e29a3bfbb0770613ffebd13ca87b9019b778a8c..b1605c7445812d2e1923ee49b9d3b15838bf980e).

PR #638, opened 2024-07-16
Add eve as owner of dependabot managed files

tcarmet:


tcarmet:
doesn't work as expected, worth the shot :')

PR #646, opened 2024-07-16
dependabot: combine patterns with exclude-pattern pip

tcarmet:


tcarmet:
not needed, went too fast on my analysis.

Issue #577, opened 2024-04-11
run user-defined functionality in `job_started` and `job_completed` hooks


harryfinbow:
We would like the ability to run some additional commands during the `job_started` and `job_completed` hooks. Currently they just run the commands defined in the `startup.sh` script which would require rebuilding the `runner-manager` image for any changes.

> I am happy to implement this myself, just wanted to get opinions as there isn't a clear "best" implementation for me to work on, hence the issue :)

## Use Case
Logging into to a private ECR repository when runners start a job.

### Example
```
function job_started {
	...

        aws ecr get-login-password | docker login

	echo "Done"
}
```

### Config Option
An additional config option (in the `runner-manager.yaml`) which would take in a string of additional commands and run them after the default commands but before the `echo "Done"`.
```
name: runner-manager
runner_groups:
- name: runner_group
  job_started: |-
    echo "Doing something"
    aws ecr get-login-password | docker login
    echo "Doing something else"
  backend:
    name: aws
    config:
      region: eu-west-1
    instance_config:
      image: ami-123456789
```
#### Limitations
Does not allow modifications to be made to the existing commands in the hooks or run commands before/in between the existing set. 

### Mount `job_started`/`job_completed` scripts
The `job_started` and `job_completed` functions could be pulled out of `startup.sh` as their own scripts so that they can be mounted, similar to how the `runner-manager.yaml` config file is, allowing for full control over these hooks.

### Mount `startup.sh` script
Similar to "Mount `job_started`/`job_completed` scripts" but mount the entire `startup.sh` script for full control over what is happening during the startup of the node.

PR #649, opened 2024-07-17
run user-defined functionality in job_started and job_completed hooks

evawoodbridge:
Adds the ability to inject bash into the job started and job completed hook through the runner group in the runner-manager.yaml config

Resolves issue raised: #577


tcarmet (commented):
This is looking pretty good! Thank you for this valuable contribution!

If possible and if it's not too much to ask, I would love to see some basic templating test, just like it is done in the [base backend tests](https://github.com/scality/runner-manager/blob/main/tests/unit/backend/test_base.py#L33). In the same spirit of what you did in [the runner_group test](https://github.com/scality/runner-manager/pull/649/files#diff-62f0ea8a06f1782f0289ed747080a9ce7634fbeb3aa5ca3202ecfeb114dd4f45), just to ensure the templating is behaving as expected,

Let me know what you think and thanks again.

tcarmet (commented):
Thanks for adding the extra test, made a small suggestion to try to make it work and we should be good to go 🚀 

Have a look at the [formatting correction needed](https://github.com/scality/runner-manager/actions/runs/9992335834/job/27627882893?pr=649) as well. If you have `trunk` installed, simply running `trunk fmt` should do the trick. If you don't have it installed, you can use the [codespace provided](https://codespaces.new/scality/runner-manager) with the repo.

tcarmet (approved):

PR #667, opened 2024-08-29
PTFE-1959 support deletion of non-existent vsphere vm

tcarmet:


typings/com/vmware/vapi/std/errors_client.pyi
nootal, line 1: you sure you want to commit this file?

nootal (approved):

typings/com/vmware/vapi/std/errors_client.pyi
tcarmet: yes, stub files are required, and if they are not available I have to generate them manually.

PR #678, opened 2024-09-11
Do not block job from running on runner hooks

tcarmet:
A softer approach to the job started hook checks that runs systematically. 
To avoid a situation where those hook might do more harm than good, if they continuously fail after a certain amount of tries, release the runner and let it do its thing.

### DIFF
--- a/runner_manager/bin/startup.sh
+++ b/runner_manager/bin/startup.sh
@@ -43,15 +43,29 @@ function job_started {
 	echo "Running preliminary checks..."
 	# Check if apt, dnf, or yum is running
 	echo "Ensure no package manager is running"
+	local tries=0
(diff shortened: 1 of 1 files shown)


nootal (approved):

liaud (approved):
Thank you! I wonder why it couldn't be started in the first place but errors should already be logged, not sure where it went through :thinking:

tcarmet:
> Thank you! I wonder why it couldn't be started in the first place but errors should already be logged, not sure where it went through 🤔

If I remember well it was stuck trying to check if docker was up. Which could be an issue later in the build, but it's better not to get the job blocked for hours due to this. Might as well fail and re-run the job.

PR #681, opened 2024-09-23
aws: allow EC2 instance access to metadata tags

cej-harris:
At the moment it is not possible to allow EC2 instances to access their metadata tags when they are created. The changes in this PR adds a field within the AWS backend to create EC2 instances, specifying if the `InstanceMetadataTags` option in `InstanceMetadataOptions` is `disabled` (the default) or `enabled`. This option can be configured with the `instance_metadata_tags` option in `runner-manager.yaml`


tcarmet (approved):
LGTM, thanks for the contribution !

PR #685, opened 2024-10-04
openstack: Implement backend and test suite

cej-harris:
Implements a backend enabling the use of an OpenStack environment to be used by `runner-manager`.

- Configured using either environment variables or `clouds.yaml` (as implemented by [`openstacksdk`](https://pypi.org/project/openstacksdk/))
- The test suite can be run either against a real OpenStack environment (by setting the `USE_LIVE_OPENSTACK` environment variable to `"true"` and configuring the parameters required by `openstacksdk`)
  - If a real OpenStack environment is not available a number of mock classes are implemented to simulate an OpenStack environment
  - `test_mock_signatures` ensures that the mock interfaces matches the interfaces provided by `openstacksdk`

I've tested this against an OpenStack environment available to me and the runners spawned function and register correctly in Github.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ coverage.xml
 config.yaml
 .env
 .reports
+clouds.yaml
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -27,6 +27,7 @@ services:
     volumes:
       - ./runner_manager:/app/runner_manager
       - ./config.yaml:/app/config.yaml
+      - ./clouds.yaml:/app/clouds.yaml
       - /var/run/docker.sock:/var/run/docker.sock
   webhook:
     profiles:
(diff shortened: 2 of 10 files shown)


cej-harris:
I think the Docker build fails because `netifaces` requires the `python3-dev` package to be installed. In testing I think I got around this by changing the base image from `python3.11-slim` to `python3.11` and forgot to include it as part of this PR. Alternatively it should be possible to install the `python3-dev` package on top of the `slim` image, but this negates some of the benefits from using the slim image.

If there's still a need for a small image size, I believe it should be possible to use a multi-stage build to install `netifaces` in the full Python base image and copy the Python environment and other required files into the `slim` Python base image.

Let me know which option you prefer and I'll have a go at implementing and testing it.

tcarmet (approved):

PR #700, opened 2025-01-16
aws-backend: add support for running instances on multiple subnets

timbrown5:
Update the AWS backend (and instance config) to add the ability to run instance on a set of subnets (and
so availability zone).

The main use-case for this is to allow instance to come up in multiple availability zones. A user would create one (or more) subnets in an availability zone and the list the subnets in the `subnet_configs` section. This has two main advantages:
- There is some crude load balancing (subnets are picked at random from the list)
- If an instance fails to start we can retry in another subnet. This will help in cases where an AZ has been starved of all the instance and instance not longer start.

To support this (without breaking existing configs) I added a new `subnet_configs` section to the instance config. This is mutually exclusive to the existing `subnet_id` section.
If `subnet_configs` is set, then the backend will choose one of the subnets at random and try to bring up the instance there. If this fails. then another subnet will be tried, until one works. If none of the subnets work, an the exception will be re-raised.

### DIFF
--- a/runner_manager/backend/aws.py
+++ b/runner_manager/backend/aws.py
@@ -92,8 +92,10 @@ class AWSBackend(BaseBackend):
         return runner
 
     def _create(self, runner: Runner, instance_resource: AwsInstance) -> Runner:
-        instance = self.client.run_instances(**instance_resource) # type: ignore
-        runner.instance_id = instance["Instances"][0]["InstanceId"] # type: ignore
+        instance = self.client.run_instances(**instance_resource)
+        # Allow this to raise exception as we don't want to track an instance that
+        # doesn't have an instance ID.
+        runner.instance_id = instance["Instances"][0]["InstanceId"]  # type: ignore
         return runner
 
     def delete(self, runner: Runner):
--- a/runner_manager/backend/aws.py
+++ b/runner_manager/backend/aws.py
@@ -1,6 +1,6 @@
 from copy import deepcopy
-from typing import List, Literal, Optional, Sequence
 from random import shuffle
+from typing import List, Literal, Optional, Sequence
 
 from boto3 import client
 from botocore.exceptions import ClientError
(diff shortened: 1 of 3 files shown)


tcarmet:
Hey @timbrown5 ! Thank you for the contribution!

Haven't got the time to review it properly yet, but will do soon, in the meantime, don't worry about codecov complaining, we don't execute those tests in the CI, but do check the linting issues reported (should be easily fixed with a `trunk fmt` if you have `trunk` installed).

Cheers!

tcarmet (approved):
just a small change request, and let's merge it.

Thanks again

runner_manager/backend/aws.py
```diff
@@ -31,15 +34,70 @@ def client(self) -> EC2Client:
...
 
+    def _create_from_subnet_config(
+        self, runner: Runner, subnet_configs: Sequence[AwsSubnetListConfig]
+    ) -> Runner:
+        # Randomize the order of the Subnets - very coarse load balancing.
+        # TODO: Skip subnets that have failed recently. Maybe with an increasing backoff.
+        order = list(range(len(subnet_configs)))
+        shuffle(order)
+        subnet_config = self.instance_config.subnet_configs
+        print(f"Order: {order}")
```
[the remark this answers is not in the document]
timbrown5: In truth, I'm not sure how useful this message is. If you're happy, I can remove it?

runner_manager/backend/aws.py
```diff
@@ -31,15 +34,68 @@ def client(self) -> EC2Client:
...
+            except Exception as e:
+                log.warn(
+                    f"Creating instance in subnet {subnet_config['subnet_id']} failed with '{e}'. Retrying with another subnet."
+                )
+                if idx >= len(order) - 1:
+                    raise e
+        return runner
+
+    def _create(self, runner: Runner, instance_resource: AwsInstance) -> Runner:
+        instance = self.client.run_instances(**instance_resource)
```
timbrown5, line 93: @tcarmet, [the linter](https://github.com/scality/runner-manager/actions/runs/12989983089/job/36239730767?pr=700#step:6:439) is complaining about this line, but it's not a addition,  just moved from the original code. Do you have a preferred way to fix it (I've silenced some warnings before, but I'm not 100% that that's the best plan).

tcarmet: Thanks for letting me know, I'm not sure how to fix it as well, but let's merge it as is, I think it's minor and it won't complain on the next lint checks.

PR #712, opened 2026-04-10
(PTFE-3162) Fix volume delete for scaleway

matthiasL-scality:
This pull request improves the reliability of deleting block storage volumes in the Scaleway backend by adding retry logic when volumes are temporarily stuck in the "in_use" state. It also adds comprehensive unit tests to ensure the new retry behavior works as expected in both success and failure scenarios.

**Improvements to Scaleway block volume deletion:**

* Added exponential backoff retry logic to the `delete` method in `scaleway.py` so that if a block storage volume is still "in_use" or marked as a "protected_resource", the system will retry deletion up to five times with increasing wait times between attempts. This handles cases where Scaleway takes time to detach volumes after server deletion. [[1]](diffhunk://#diff-ad3a8001c753a3db1894703fb801fbb1a9a010b659c4d24050ba74c194035cffR467-R471) [[2]](diffhunk://#diff-ad3a8001c753a3db1894703fb801fbb1a9a010b659c4d24050ba74c194035cffR480-R504)
* Ensured that if all retries are exhausted and the volume still cannot be deleted, an appropriate warning is logged and the process continues gracefully. [[1]](diffhunk://#diff-ad3a8001c753a3db1894703fb801fbb1a9a010b659c4d24050ba74c194035cffR480-R504) [[2]](diffhunk://#diff-ad3a8001c753a3db1894703fb801fbb1a9a010b659c4d24050ba74c194035cffL492-L493)

**Testing enhancements:**

* Added `test_delete_volume_in_use_retry_succeeds` to verify that the retry logic successfully deletes a volume after temporary "in_use" errors.
* Added `test_delete_volume_in_use_retry_exhausted` to verify that a warning is logged and the correct result is returned if the volume remains "in_use" after all retries.

### DIFF
--- a/runner_manager/backend/scaleway.py
+++ b/runner_manager/backend/scaleway.py
@@ -18,6 +18,7 @@ from scaleway.instance.v1 import (
 )
 from scaleway.instance.v1.custom_api import InstanceUtilsV1API
 from scaleway.marketplace.v2 import MarketplaceV2API
+from scaleway_core.api import ScalewayException
 
 from runner_manager.backend.base import BaseBackend
 from runner_manager.models.backend import (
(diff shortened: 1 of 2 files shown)


runner_manager/backend/scaleway.py
```diff
@@ -464,33 +464,54 @@ def delete(self, runner: Runner) -> int:
...
-                                f"Instance volume {volume_id} deleted successfully"
+                                f"Block storage volume {volume_id} deleted successfully"
                             )
-                        else:
-                            raise block_error
+                            deleted = True
+                            break
+                        except Exception as block_error:
+                            block_error_msg = str(block_error)
+                            if (
```
cmeron: Instead of casting the Exception object, you can use:

- `block_error.status_code` which is the HTTP exit code as int()
- `block_error.response` which is the string of the error

I didn't test it in this specific case but it works in other Scaleway API calls

runner_manager/backend/scaleway.py
```diff
@@ -464,33 +464,54 @@ def delete(self, runner: Runner) -> int:
...
+                                "404" in block_error_msg
+                                or "not_found" in block_error_msg.lower()
+                            ):
+                                # Not a block storage volume, fall through to Instance API
+                                break
+                            if (
+                                "in_use" in block_error_msg.lower()
+                                or "protected_resource" in block_error_msg.lower()
+                            ) and attempt < max_attempts - 1:
```
cmeron: It seems that if max_attempts is set to 1 it will never run 
(not a big deal though)

cmeron (approved):
Please take a look at my comments but overall LGTM!

matthiasL-scality:
@claude[agent] review

